PR #26, opened 2024-04-02
chore(DEV-10044): update Tanstack Query to v5

hypsteria:
Migrating to TanStack Query v5
Source: https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5


packages/sdk-react/src/components/products/ProductDeleteModal/ProductDeleteModal.tsx
```diff
@@ -65,7 +65,7 @@ export const ProductDeleteModal = ({
         <Button
           variant="outlined"
           color="error"
-          disabled={deleteProductMutation.isLoading || isInitialLoading}
+          disabled={deleteProductMutation.isPending || isInitialLoading}
```
radist2s, line 68: `isInitialLoading` is [deprecated](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#:~:text=is%20deprecated%20now%20and%20will%20be%20removed%20in%20the%20next%20major%20version). Would be cool to remove its usage globally.

packages/sdk-react/src/core/context/MoniteProvider.tsx
```diff
@@ -135,6 +135,10 @@ const createQueryClient = (i18n: I18n, sentryHub: Hub | undefined) =>
 
             return;
           }
+
+          toast.error(message, {
```
radist2s, line 139: Didn't we decide it was better not to handle all errors in this manner? Maybe I'm missing something in the implementation.

packages/sdk-react/src/components/products/ProductDeleteModal/ProductDeleteModal.tsx
hypsteria: fixed 

I saved the logic according to tanstack query v5 guide
```
isLoading -> isPending
isInitialLoading -> isLoading
```
image

packages/sdk-react/src/core/context/MoniteProvider.tsx
hypsteria: It was missed part of the code. You can see that it is already implemented for **mutationCache**. I added it for **queryCache**.

We are handling onError event in one place instead of handling it in each query. **onError** callback was deprecated in v5 https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#callbacks-on-usequery-and-queryobserver-have-been-removed

nekedos: As I remember, we decided to remove all onError and onSuccess callbacks from all queries and mutations.

packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/ApprovalPolicyDetails.tsx
```diff
@@ -39,11 +39,11 @@ export const ApprovalPolicyDetails = ({
 
   const {
     isLoading,
-    isInitialLoading,
+    isPending,
     data: approvalPolicy,
   } = useApprovalPolicyById(id);
 
-  if (id && (isLoading || isInitialLoading)) {
+  if (id && isLoading && isPending) {
```
radist2s, line 46: Probably, `if (id && isLoading)` would be enough for the case.

packages/sdk-react/src/components/onboarding/OnboardingFile/OnboardingFileUploader/OnboardingFileUploader.tsx
```diff
@@ -71,17 +71,17 @@ export const OnboardingFileUploader = ({
...
               <CircularProgress color="inherit" size={22} />
             ) : (
               <CloudUpload />
             )
           }
         >
-          {isLoading && t(i18n)`Processing...`}
-          {!isLoading && t(i18n)`Choose document`}
+          {isPending && t(i18n)`Processing...`}
+          {!isPending && t(i18n)`Choose document`}
```
radist2s, line 84: `isPending` will be false when `isError` is `true`, would be better to use `isSuccess` instead

packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/ApprovalPolicyDetails.tsx
hypsteria: left only `isLoading ` here

packages/sdk-react/src/core/context/MoniteProvider.tsx
hypsteria: fixed

Removed global error handling.

Placed **useEffect** with error handling in places for **useQueries** that had it before migration to v5.

radist2s (approved):
Monumental! Grandiose! The whole plate of pies is yours!

PR #34, opened 2024-04-11
fix(DEV-10434): fix initial loading of products in the receivables

nekedos:
A pagination_token was set as '0', which is incorrect. Replaced with an empty string.

### DIFF
--- a/packages/sdk-react/src/core/queries/useProducts.ts
+++ b/packages/sdk-react/src/core/queries/useProducts.ts
@@ -7,7 +7,6 @@ import {
   ProductServicePaginationResponse,
   ProductServiceRequest,
   ProductServiceResponse,
-  ProductsService,
   ProductsServiceGetAllRequest,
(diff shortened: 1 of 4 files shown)


packages/sdk-react/src/core/queries/useProducts.ts
```diff
@@ -87,7 +86,7 @@ export const useInfiniteProducts = (
         paginationToken: pageParam,
       });
     },
-    initialPageParam: '0',
+    initialPageParam: '',
```
radist2s, line 89: Why not make `initialPageParam` as `undefined`?

nekedos: Thanks, I will check

nekedos: I checked. Schema requires a string.

radist2s (approved):

PR #39, opened 2024-04-15
feat(DEV-9561): theme switcher

saladNights:
Themes were extracted to package to reuse in sdk-react and sdk-demo.

Theme switcher was added to demo for dev environment only.

Screenshot 2024-04-15 at 12 12 41


packages/sdk-demo/src/components/Layout/index.tsx
```diff
@@ -73,12 +75,21 @@ export const DefaultLayout = ({ children, siderProps }: DefaultLayoutProps) => {
...
-              mb: 2,
+              p: 2,
             }}
           >
-            {siderProps?.footer}
+            <Stack direction="column" spacing={2} ml={2}>
+              {/*Themes are unfinished.*/}
+              {/*We want to show the theme switcher only in development mode and on the dev deployment only.*/}
+              {(process.env.NODE_ENV === 'development' ||
+                location.pathname.indexOf('dev.monite.') > 0) && (
```
radist2s, line 88: You probably where thinking of `location.hostname`, and its usage with 
`location.hostname.indexOf('dev.monite.')`.
Btw, it is always better to check, is the `location` really exists (the code could run not only on in the browser): `typeof location !== 'unefined'`

MadMed677, line 88: I'd suggest moving the condition into a variable with a name, otherwise, it might be a little bit hard to understand what we're checking here.

[As an example of how it's done for Sentry checks](https://github.com/team-monite/monite-sdk/blob/2504752c8b6ea7c999e0a3fa8754cb74e3df49e9/packages/sdk-react/src/core/services/Sentry.factory.ts#L38)

nekedos: It would be better to use `api_url` from `/config.json` to check stand instead of locale.
Also, we can add a `stand` to `config.json` to avoid parsing strings.

packages/sdk-themes/src/themes/monite.ts
```diff
@@ -0,0 +1,203 @@
+import { grey } from '@mui/material/colors';
+import { Components } from '@mui/material/styles/components';
+import { Palette, PaletteOptions } from '@mui/material/styles/createPalette';
+import { Theme, ThemeOptions } from '@mui/material/styles/createTheme';
+import { TypographyOptions } from '@mui/material/styles/createTypography';
+import type {} from '@mui/x-data-grid/themeAugmentation';
```
nekedos, line 6: Is it correct?

saladNights: yes https://mui.com/material-ui/about-the-lab/#typescript

packages/sdk-demo/src/components/Layout/index.tsx
saladNights: fixed

saladNights: fixed

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
```diff
@@ -0,0 +1,102 @@
...
+};
+
+export const ThemeSelector = () => {
+  const { i18n } = useLingui();
+  const { root } = useRootElements();
+
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+
+  const { themeIndex, colorMode, setTheme, toggleColorMode } =
```
radist2s, line 40: Let's move usage of `setTheme()` on the top of `ThemeSelector`, to make it compatible with Next.js:

```tsx
<ThemeSelector onChange={setTheme} />
```

packages/sdk-demo/src/components/Layout/index.tsx
```diff
@@ -22,15 +25,32 @@ type DefaultLayoutProps = {
 export const DefaultLayout = ({ children, siderProps }: DefaultLayoutProps) => {
   const location = useLocation();
   const { data: user } = useEntityUserByAuthToken();
+  const [config, setConfig] = useState<ConfigSchema | null>(null);
   const [pagePadding, setPagePadding] = useState(4);
 
-  React.useEffect(() => {
+  useEffect(() => {
+    const fetchConfig = async () => {
```
radist2s, line 32: It's time, to use the full power is Suspense! We already used this [approach](https://github.com/team-monite/monite-sdk/blob/0ccc9f84afc5a60351dcc90186b98746c68bb7e9/packages/sdk-react/src/core/context/I18nLocaleProvider.tsx#L97) to compile translation messages in runtime.

Brief implementation:

```tsx
import React, { lazy, memo, Suspense } from 'react';

const ConfigProvider = memo(lazy(async () => {
  const config = await getConfig();
  return {
    default: function ConfigProviderLoader({ children }: { children: React.ReactNode }) {
      return (
        <ConfigContext.Provider value={config}>
          {children}
        </ConfigContext.Provider>
      )
    }
  }
}));

const 

```

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
radist2s: Btw, I changed my mind. We will wrap Next.js App with the same Provider and themes list from `@team-monite/themes`, and reuse the logic.

packages/sdk-demo/src/context/themeContext/themeContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+  colorMode: 'light',
+  setTheme: () => {},
+  toggleColorMode: () => {},
+});
+
+export const ThemeContextProvider = ({
+  children,
+}: ThemeContextProviderProps) => {
+  const [localStorageTheme, setLocalStorageTheme] =
+    useLocalStorage<LocalStorageTheme>('theme');
```
radist2s, line 58: A few notes:
- `useLocalStorage(key, initialData)` has `initialData` argument
- Next.js can't access Local Storage during SSR. Probably, we gonna make server-side requests to retrieve metadata with the theme config from Clerk's private meta data (why not?)

We need to add custom handlers to the provider:

```tsx
<ThemeContextProvider
  onChange={save_To_Local_Storage_Or_To_The_Clerk}
  value={initialTheme_FromLocal_Storage_or_Backend_Request}
/>
```

packages/sdk-themes/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@team-monite/sdk-themes",
+  "version": "0.1.0",
+  "private": true,
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "scripts": {
+    "dev": "yarn clean && tsc --watch",
+    "build": "yarn clean && tsc",
+    "clean": "rm -rf dist"
```
radist2s, line 10: Please, add `rimraf`, it's more convenient

packages/sdk-themes/src/themes/monite.ts
radist2s: I worry that this will also make a change in the customers' Types? Will this augmentation only affect _our theme_ types, or will customer types be augmented as well?

saladNights: since `@mui/x-data-grid/themeAugmentation` uses `declare module`, not `declare global`, I suppose it should affect only the current theme.

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
saladNights: ok

packages/sdk-demo/src/components/Layout/index.tsx
saladNights: the config context created

packages/sdk-themes/package.json
saladNights: fixed

packages/sdk-demo/src/context/themeContext/themeContext.tsx
saladNights: added

packages/sdk-react/src/core/context/MoniteProvider.tsx
```diff
@@ -232,9 +231,7 @@ export const MoniteProvider = ({
         key={moniteInstanceKey}
         sentryHub={sentryHub}
       >
-        <MoniteThemeContext.Provider
-          value={theme ?? defaultMoniteLightThemeOptions}
-        >
+        <MoniteThemeContext.Provider value={theme}>
```
MadMed677, line 234: If our user migrates to a new version and doesn't pass any `theme` the theme will be set to a Material, won't it?
If it's true isn't it a major release?

saladNights: yes, you are right.
I discussed it with Nikita. We agreed that it is not a breaking change, everything will work, just the UI will look a little bit different.
Probably all our current customers use their own theme.

packages/sdk-demo/src/context/configContext/configContext.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { createContext, useContext, memo } from 'react';
+
+import { getConfig, ConfigSchema } from '@/core/getConfig';
+
+type ConfigContextType = ConfigSchema | null;
+
+const ConfigContext = createContext<ConfigContextType>(null);
```
radist2s, line 7: Don't mind renaming the file to `ConfigContext/ConfigContext.tsx`? I would also ask to check the `themeContext` too.

packages/sdk-demo/src/context/configContext/configContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface ConfigProviderProps {
+  config: ConfigSchema;
+  children: React.ReactNode;
+}
+
+const ConfigProviderComponent = memo(
```
radist2s, line 14: `memo(...)` is not needed here, as `config` and `children` are passed directly to the context. `memo` is needed to optimize the component's logic computing.

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+        }
+        aria-haspopup="true"
+        aria-expanded={isMenuOpen ? 'true' : undefined}
+        aria-label="theme-menu-button"
+        variant="outlined"
+        onClick={() => setIsMenuOpen(!isMenuOpen)}
+      >
+        {getButtonLabel(themeIndex, i18n)}
+        <Menu
+          id="theme-selector"
```
radist2s, line 69: Probably, `useId(...)` result should be used instead

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+};
+
+interface ThemeSelectorProps {
+  themeConfig: ThemeConfig;
+  onChange: (themeConfig: ThemeConfig) => void;
+}
+
+export const ThemeSelector = ({
+  themeConfig,
```
radist2s, line 39: - I would recommend renaming `themeConfig` to `value`, as it would be conventional to _Select_.
- I would prefer to keep MUI naming for the components, so, _ThemeSelector_ probably may be renamed to _ThemeSelect_
- Just Brilliant, btw 🔷

packages/sdk-demo/src/hooks/useTheme.tsx
```diff
@@ -0,0 +1,36 @@
...
+  };
+  const [themeConfig, setThemeConfig] = useLocalStorage<ThemeConfig>(
+    'themeConfig',
+    defaultThemeConfig
+  );
+
+  return {
+    theme: getTheme(themeConfig || defaultThemeConfig),
+    themeConfig: themeConfig || defaultThemeConfig,
+    setThemeConfig,
```
radist2s, line 34: What if we simplified the returned data:
```
return {
     theme: themeConfig || defaultThemeConfig,
     setTheme: setThemeConfig,
}
```

Then, `getTheme(...)` could be used both in Next.js and in the SDK Demo where it is needed:
```tsx
<AppMoniteProvider theme={getTheme(theme)} />
```

packages/sdk-demo/src/types/index.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface ThemeConfig {
+  themeIndex: 'material' | 'monite';
```
radist2s, line 2: I would like to name props according to [MUI naming convention](https://mui.com/material-ui/customization/dark-mode/#dark-mode-by-default):
```
themeIndex => variant | palette
colorMode   => mode
```

packages/sdk-demo/src/context/configContext/configContext.tsx
saladNights: fixed

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
saladNights: fixed

packages/sdk-demo/src/components/Layout/ThemeSelector.tsx
saladNights: good point, thanks. fixed

packages/sdk-demo/src/hooks/useTheme.tsx
saladNights: fixed

packages/sdk-demo/src/types/index.d.ts
saladNights: fixed. I chose 'variant' because 'palette' is just a part of the theme options object. 'variant' is more general

packages/sdk-react/src/core/context/MoniteProvider.tsx
MadMed677: Hm, okay. I hope so

radist2s (approved):
💯

PR #43, opened 2024-04-18
feat(DEV-9977): update msw from 0.*.* to 2.2.*

MadMed677:
Migrated into a new major version of a mock service worker.
So I don't think that you have to review all files only a few will be enough from my point of view.


packages/sdk-react/src/mocks/utils.ts
```diff
@@ -13,14 +12,12 @@ export const getMockPagination = (
   return { prevPage, nextPage };
 };
 
-export const delay = (duration?: number): ResponseTransformer => {
+export const delay = (duration?: number): Promise<void> => {
   if (process.env.TESTS) {
```
radist2s, line 16: I doubt it, do we need a custom `dealy(...)` implementation? According to [MSW documentation](https://mswjs.io/docs/api/delay/#implicit-delay), the native `delay` must already be skipped during the tests:

> Note that when testing in Node.js, the implicit delay is negated to prevent it affecting the test performance. If you wish to delay responses in Node.js, please use one of the options below.

MadMed677: Hello, @radist2s, I am returning to you with my research.
As I've tested it out it doesn't work but, I suppose, not because MSW doesn't work well I think it's related to our tests. As I can see MSW default node response still has timeout but it's short (5ms) ([github link](https://github.com/mswjs/msw/blob/main/src/core/delay.ts#L6))

Our tests right now rely on immediate response timeout and because of that sometimes tests are flucky. Let me provide some examples:
Screenshot 2024-04-23 at 11 47 07

In that test, we rely on the that `prev` and `next` buttons already have correct attributes (`disabled`) but in reality sometimes our buttons are still not available until:
1. Buttons also handle mutations. So the button won't be available until the mutation is not loaded (*mutation.isLoading)
2. Or the interface does not render immediately and it renders on the next tick and when MSW sets timeout to 5ms it postpones response **after** render. In our current approach when we return everything immediately we return data **before** render

So I see 2 options right here:
1. Keep using our approach with immediate response (it may lead to some flucky tests in the future or even right now)
2. Create a separate task to rewrite tests to be more error-prone and handle situations when data may respond in 5ms or even more

What do you think? I hope I managed to explain but if not feel free to text me back

radist2s: @MadMed677, Let's keep our custom approach, but it would be great to clarify with the MSW developers why, in their opinion, a 5ms delay is needed, even for tests.

radist2s (approved):
💪🔩

PR #48, opened 2024-04-23
chore(proposal): improve `<CounterpartBankForm/>` tests stability 🟢

radist2s:
- Replaced unstable `await waitUntilTableIsLoaded()` with specific Query waiter `checkCounterpartQueriesLoaded(...)`
- Increased timeouts time


packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartBankForm/CounterpartBankForm.test.tsx
```diff
@@ -37,7 +39,9 @@ describe('CounterpartBankForm', () => {
       </MoniteStyleProvider>
     );
 
-    await waitUntilTableIsLoaded();
+    await waitFor(() => checkCounterpartQueriesLoaded(testQueryClient), {
```
MadMed677, line 42: Hello. I do have 2 questions:
1. Why `waitUntilTableIsLoaded` is unstable?
2. Default timeout is 1 second. Increasing timeout to 5 or even 10 seconds looks like hack and if everything has been written correct timeout to 1 second should be more than enough

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartBankForm/CounterpartBankForm.test.tsx
```diff
@@ -346,7 +362,28 @@ describe('CounterpartBankForm', () => {
...
     });
   });
 });
+
+/**
+ * Check if all counterpart queries are not fetching and do not have an error.
+ *
+ * @param queryClient
+ */
+function checkCounterpartQueriesLoaded(queryClient: QueryClient) {
```
MadMed677, line 375: I see a drawback here. Right now we check queries but before we checked the `spinner`. From my point of view current approach is less stable because if we add another check to show the bank form the test will fail but before we just wait until the spinner disappears and it doesn't matter which checks we have.

Overall don't get the point to rewrite it into specific query checks

radist2s: In my opinion, the correct test would look like this:
- Test whether a spinner is displayed to show the load status, depending on the `status` of this or that `Query`
- Other tests, the success of which depends on the `status` of this or that `Query`.

Furthermore, if the test crashes because `waitUntilTableIsLoaded(...)` failed to complete successfully, there is no way to know which query is _stuck_ or crashed. With the `checkCounterpartQueriesLoaded(...)` implementation, it is visible _what the problem is_, and _which query is problematic_.

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartBankForm/CounterpartBankForm.test.tsx
MadMed677: > For example, there may be 2 spinners appearing sequentially on screen (as in Super Components, for example).

I dunno, for me it's a mistake if we have 2 or more spinners sequentially. I just curious why in other tests `waitUntilTableIsLoaded` works but in `CounterpartBankForm` isn't.

From my point of view there is some mistake inside `CounterpartBankForm` and it's not related to `waitUntilTableIsLoaded` otherwise all our tests will fail but right now only this one fails. So I'd like to check what have been changed inside the logic why it's so error-prone

nekedos: In my opinion, it is not necessary to use `waitUntilTableIsLoaded `.

We can use this code instead:
```ts
const submitBtn = await waitFor(
  () =>
    screen.getByRole('button', {
      name: t`Add bank account`,
    })
);

nekedos (approved):

radist2s:
- Removed `checkCounterpartQueriesLoaded(...)` with target elements search inside `waitFor(...)`
- Removed timeouts

MadMed677 (approved):

PR #53, opened 2024-05-02
feat(DEV-4699): Provide an ability to edit an invoice

MadMed677:
ar-edit

Works on `dev` stand

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -128,7 +128,7 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
               key="next"
               color="primary"
               type="submit"
-              form="receivablesDetailsForm"
+              form="Monite-Form-receivablesDetailsForm"
               disabled={createReceivable.isPending}
             >{t(i18n)`Create`}</Button>
           </Box>
@@ -138,7 +138,7 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
       <DialogContent>
         <FormProvider {...methods}>
           <form
-            id="receivablesDetailsForm"
+            id="Monite-Form-receivablesDetailsForm"
             noValidate
             onSubmit={handleSubmit((values) => {
               if (values.type !== InvoiceResponsePayload.type.INVOICE) {
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useId, useMemo, useState } from 'react';
 import { useForm, FormProvider } from 'react-hook-form';
 
 import { useDialog } from '@/components';
@@ -104,12 +104,12 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
     CurrencyEnum | undefined
   >(settings?.currency?.default);
 
+  // eslint-disable-next-line lingui/no-unlocalized-strings
+  const formName = `Monite-Form-receivablesDetailsForm-${useId()}`;
+
   if (isSettingsLoading) {
     return <LoadingPage />;
   }
-
-  // eslint-disable-next-line lingui/no-unlocalized-strings
-  const formName = 'Monite-Form-receivablesDetailsForm';
 
   return (
     <>
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -108,6 +108,9 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
     return <LoadingPage />;
   }
 
+  // eslint-disable-next-line lingui/no-unlocalized-strings
+  const formName = 'Monite-Form-receivablesDetailsForm';
+
   return (
     <>
       <DialogTitle>
@@ -128,7 +131,7 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
               key="next"
               color="primary"
               type="submit"
-              form="Monite-Form-receivablesDetailsForm"
+              form={formName}
               disabled={createReceivable.isPending}
             >{t(i18n)`Create`}</Button>
           </Box>
@@ -138,7 +141,7 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
       <DialogContent>
         <FormProvider {...methods}>
           <form
-            id="Monite-Form-receivablesDetailsForm"
+            id={formName}
             noValidate
             onSubmit={handleSubmit((values) => {
               if (values.type !== InvoiceResponsePayload.type.INVOICE) {
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -37,7 +37,7 @@ import { ItemsSection } from './sections/ItemsSection';
 import { PaymentSection } from './sections/PaymentSection';
 import {
   getCreateInvoiceValidationSchema,
-  ICreateReceivablesForm,
+  CreateReceivablesFormProps,
 } from './validation';
 
 enum ReceivableViewEnum {
@@ -67,7 +67,7 @@ export const CreateReceivables = (props: InvoiceDetailsCreateProps) => (
 const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
   const { i18n } = useLingui();
   const dialogContext = useDialog();
-  const methods = useForm<ICreateReceivablesForm>({
+  const methods = useForm<CreateReceivablesFormProps>({
     resolver: yupResolver(getCreateInvoiceValidationSchema(i18n)),
     defaultValues: useMemo(
       () => ({
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { Controller, useFormContext } from 'react-hook-form';
 
 import { CountryInvoiceOption } from '@/components/receivables/InvoiceDetails/CreateReceivable/components/CountryInvoiceOption';
@@ -35,7 +35,24 @@ const detailsGridItemProps = {
   lg: 3,
 };
 
-export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
+/** All available fields for current component */
+const allFields = [
+  'entity_vat_id_id',
+  'tax_id',
+  'fulfillment_date',
+  'purchase_order',
+] as const;
+
+interface IEntitySectionProps extends ISectionGeneralProps {
+  /**
+   * Describes which fields should be hidden from the user
+   *  (example: `purchaise_order` is not available for editing
+   *   and we want to hide it. But for [CREATE] request it should be sent)
+   */
+  hidden?: ['purchase_order'];
+}
+
+export const EntitySection = ({ disabled, hidden }: IEntitySectionProps) => {
   const { i18n } = useLingui();
   const { control, watch, resetField, setValue } =
     useFormContext<ICreateReceivablesForm>();
@@ -50,13 +67,31 @@ export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
   const [isSameAsInvoiceDateChecked, setIsSameAsInvoiceDateChecked] =
     useState<boolean>(false);
 
+  const visibleFields = useMemo(() => {
+    if (!hidden) {
+      return allFields;
+    }
+
+    return allFields.filter((field) => field !== 'purchase_order');
+  }, [hidden]);
+
+  const gridItemProps = useMemo(() => {
+    const proportion = Math.floor(12 / visibleFields.length);
+
+    return {
+      ...detailsGridItemProps,
+      md: proportion,
+      lg: proportion,
+    };
+  }, [visibleFields]);
+
   return (
     <Stack spacing={1}>
       <Typography variant="subtitle2">{t(i18n)`Details`}</Typography>
       <Card variant="outlined" sx={{ borderRadius: 2 }}>
         <CardContent>
           <Grid container spacing={3}>
-            <Grid item {...detailsGridItemProps}>
+            <Grid item {...gridItemProps}>
               <Controller
                 name="entity_vat_id_id"
                 control={control}
@@ -95,7 +130,7 @@ export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
                 )}
               />
             </Grid>
-            <Grid item {...detailsGridItemProps}>
+            <Grid item {...gridItemProps}>
               <TextField
                 disabled
                 fullWidth
@@ -114,7 +149,7 @@ export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
                 <FormHelperText>{t(i18n)`No TAX ID available`}</FormHelperText>
               </Collapse>
             </Grid>
-            <Grid item {...detailsGridItemProps}>
+            <Grid item {...gridItemProps}>
               <Controller
                 name="fulfillment_date"
                 control={control}
@@ -183,23 +218,25 @@ export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
                 )}
               />
             </Grid>
-            <Grid item {...detailsGridItemProps}>
-              <Controller
-                name="purchase_order"
-                control={control}
-                render={({ field, fieldState: { error } }) => (
-                  <TextField
-                    fullWidth
-                    variant="outlined"
-                    label={t(i18n)`Purchase order`}
-                    error={Boolean(error)}
-                    helperText={error?.message}
-                    disabled={disabled}
-                    {...field}
-                  />
-                )}
-              />
-            </Grid>
+            {visibleFields.includes('purchase_order') && (
+              <Grid item {...gridItemProps}>
+                <Controller
+                  name="purchase_order"
+                  control={control}
+                  render={({ field, fieldState: { error } }) => (
+                    <TextField
+                      fullWidth
+                      variant="outlined"
+                      label={t(i18n)`Purchase order`}
+                      error={Boolean(error)}
+                      helperText={error?.message}
+                      disabled={disabled}
+                      {...field}
+                    />
+                  )}
+                />
+              </Grid>
+            )}
           </Grid>
         </CardContent>
       </Card>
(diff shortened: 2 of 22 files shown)


packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
```diff
@@ -35,7 +35,24 @@ const detailsGridItemProps = {
...
-export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
+/** All available fields for current component */
+const allFields = [
+  'entity_vat_id_id',
+  'tax_id',
+  'fulfillment_date',
+  'purchase_order',
+] as const;
+
+interface IEntitySectionProps extends ISectionGeneralProps {
```
radist2s, line 46: Maybe we should follow the MUI style of naming interfaces?
`IEntitySectionProps` => `EntitySectionProps`

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
```diff
@@ -133,16 +161,26 @@ export function useExistingInvoiceDetails({
     );
   }, [deliveryMethod, receivable?.status]);
 
+  const isEditButtonVisible = useMemo(() => {
+    return (
```
radist2s, line 165: Probably, we have to check the user's permissions too.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -0,0 +1,287 @@
...
+              disabled={isLoading}
+            >{t(i18n)`Update`}</Button>
+          </Box>
+        </Toolbar>
+      </DialogTitle>
+      <Divider />
+      <DialogContent>
+        <FormProvider {...methods}>
+          <form
+            id="receivablesDetailsForm"
```
radist2s, line 182: It would be better to replace `receivablesDetailsForm` with the unique ID:
```ts
const formId = `Monite-Form-${useId()}`
```

Probably, you can add similar changes to [EditInvoiceDetails](https://github.com/team-monite/monite-sdk/blob/9ceb783e8a09c1b63355e25675b34fade03f16fe/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx#L182)

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
```diff
@@ -221,6 +235,14 @@ export const ExistingInvoiceDetails = (
                     disabled={loading}
                   >{t(i18n)`Delete`}</Button>
                 )}
+                {buttons.isEditButtonVisible && (
+                  <Button
```
radist2s, line 239: Alternatively, you can add permissions checking right here. Btw, _Delete_ action is missing permission checking too. Check out [an example](https://github.com/team-monite/monite-sdk/commit/bde068d520a604abf069d910e48711a36fe58039#diff-4b85029f45ba03ebdfa44b46e4f56a24dfa7c864daac70d2aaeafc61b6f1faeaR72-R76)!

radist2s:
The same cool dialog will not be shown, if you press <kbd>Esc</kbd> 😞

image

radist2s:
I think I'm confused.

- In the first screenshot we see VAT ID `27088af0`

  image

- In the next VAT ID is empty but `27088af0` refers to Tax ID.

  image

radist2s:
If update an Invoice, you will be returned to the Invoice Details dialog, but PDF will not load immediately, as it is not yet ready on the server. Is possible to specify somehow better information about PDF status? Also, the alignment of the message seems broken.

image

radist2s:
@MadMed677, huge work! Thank you!

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -0,0 +1,287 @@
...
+                       * There is no `time` in request, only year, month and date
+                       */
+                      format(values.fulfillment_date, 'yyyy-MM-dd')
+                    : undefined,
+                  /** !!! Note !!! Backend is not supported to edit `purchase_order` so we have to remove it */
+                  // purchase_order: values.purchase_order || undefined,
+                },
+              };
+
+              updateReceivableLineItems.mutate(lineItems, {
```
radist2s, line 232: If possible do not update `lineItems` if not changed, that would be cool.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -0,0 +1,287 @@
...
+                  entity_vat_id_id: values.entity_vat_id_id || undefined,
+                  fulfillment_date: values.fulfillment_date
+                    ? /**
+                       * We have to change the date as Backend accepts it.
+                       * There is no `time` in request, only year, month and date
+                       */
+                      format(values.fulfillment_date, 'yyyy-MM-dd')
+                    : undefined,
+                  /** !!! Note !!! Backend is not supported to edit `purchase_order` so we have to remove it */
+                  // purchase_order: values.purchase_order || undefined,
```
radist2s, line 228: Is the backend going to update the OpenAPI Document according to this issue?

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
```diff
@@ -50,13 +67,31 @@ export const EntitySection = ({ disabled }: ISectionGeneralProps) => {
...
 
+  const visibleFields = useMemo(() => {
+    if (!hidden) {
+      return allFields;
+    }
+
+    return allFields.filter((field) => field !== 'purchase_order');
+  }, [hidden]);
+
+  const gridItemProps = useMemo(() => {
```
radist2s, line 78: 💋

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
MadMed677: We don't handle permissions for receivables (at all) right now, unfortunately :(

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: Hm, but why? We always use specific form ids

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: Yep, it sounds good. But do you know how to check it? Because as I know `react-hook-form` will create a new array of `line_items` and if we check `current.lineItems === server.lineItems` we will have always `false` because it will be a new array.

So the only one way is to check it by deep check (check all fields by going through all items) which I think will be worse than current one.

Maybe if we could handle `isDirty` but not for the whole form but only for `line_items` we could avoid any checks but I didn't find any options to do so

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: They already have it :)

I just realized that during implementation. They have to fix it and support these fields but we never know when

MadMed677:
> the Dialog will also be displayed (as if the form becomes dirty with no real changes).

Yep, but it's impossible to fix it :( That's how react-hook-form works. It sets `isDirty: true` when you change anything and even if it's the same value react-hook-form doesn't handle it

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
radist2s: Do you mean there is no support on the backend for differentiating permissions?

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
radist2s: There may be a situation the `id` on the customer's page will conflict with the non-unique `id` of `receivablesDetailsForm`

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
MadMed677: On BE side it's supported I guess, but on FE side there is no implementation of it

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: Let's add this topic to FE sync. Because it's a general change and if it really possible to have any problem with it we could change it but globally for all forms then

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
MadMed677: Hm, why? :)
We always use these options. How is our code style related to MUI code style?

MadMed677:
> The same cool dialog will not be shown, if you press Esc 😞
> 
> image
> **Btw**, even if I didn't change anything, ~but just expanded the Select and closed it,~ the Dialog will also be displayed (as if the form becomes dirty with no real changes).

I don't know how it should work. Undortunately design team doesn't support it. Right now `ECS` works only when you want to close pop-up but current page is not a pop-up window just another _route_ to edit an invoice. If you check any other entity (payables, counterparts, etc) they will work the same

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
radist2s: I don't recall us adopting any convention, accordingly, we follow the MUI style.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
radist2s: It seems better to avoid obvious mistakes in advance than to correct them in the future.

nekedos: @MadMed677, By the way, why do we need to use `id`?

MadMed677: @nekedos When we submit the form we have to provide which form we want to submit. 
Example: https://github.com/team-monite/monite-sdk/blob/61ee304aa3066640b0a49c78d735cbdf0ac4d1b5/packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/ApprovalPolicyDetailsForm/ApprovalPolicyDetailsForm.tsx#L274

MadMed677:
> I think I'm confused.
> 
> * In the first screenshot we see VAT ID `27088af0`
>   image
> * In the next VAT ID is empty but `27088af0` refers to Tax ID.
>   image
> * Finally, updated VAT ID
>   image

Updated

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: @radist2s updated

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
MadMed677: We don't support permissions right now

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
MadMed677: I don't see any convention on our side. I personally use this format for all interfaces

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
```diff
@@ -108,6 +108,9 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
     return <LoadingPage />;
   }
 
+  // eslint-disable-next-line lingui/no-unlocalized-strings
+  const formName = 'Monite-Form-receivablesDetailsForm';
```
radist2s, line 112: Could you please refactor it with the `Monite-Form-receivablesDetailsForm-useId()`?

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
nekedos: @MadMed677 Can I personally ask you to avoid such naming and follow the MUI style?

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
radist2s: Not, sure, but probably [this is it](https://github.com/team-monite/monite-sdk/blob/bde068d520a604abf069d910e48711a36fe58039/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingReceivableDetails.tsx#L77-L81).

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
radist2s: I would prefer to keep MUI style in all naming.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const [actualCurrency, setActualCurrency] = useState(invoice.currency);
+
+  const {
+    handleSubmit,
+    setValue,
+    formState: { isDirty },
+  } = methods;
+
+  useEffect(() => {
+    setValue('default_shipping_address_id', counterpartShippingAddress?.id);
```
radist2s, line 132: I suppose, this is the reason why the form state is always `touched`. It would be better to avoid using such hacks.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/EntitySection.tsx
MadMed677: Updated

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
MadMed677: `ExistingReceivableDetails` works only for `Quotes` & `Credit Notes` it has outdated UI and have been made more than a year ago and shouldn't be supported
https://github.com/team-monite/monite-sdk/blob/1cba52101f1691de8f7027b27da07e260b6b9147/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingReceivableDetails.tsx#L60

All improvements are made only for `Invoices`. Currently `Invoices` doesn't support backend permissions

I don't want to add to this task which is called `Provide an ability to edit invoice` functionality to check permissions because it's out of scope of this task.

I created a task to support BE permissions for a new component ([task link](https://monite.atlassian.net/browse/DEV-10818)). Will it work for you in that case? To make it in a separate task

radist2s: Sure, btw, which actions are not supported?

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -0,0 +1,289 @@
...
+  }, [isDirty, onCancel]);
+
+  const updateReceivableLineItems = useUpdateReceivableLineItems(invoice.id);
+  const updateReceivable = useUpdateReceivable(invoice.id);
+
+  const isLoading =
+    updateReceivableLineItems.isPending || updateReceivable.isPending;
+
+  // eslint-disable-next-line lingui/no-unlocalized-strings
+  const formName = 'Monite-Form-receivablesDetailsForm';
```
radist2s, line 155: Would be better to refactor as

```
const formName = `Monite-Form-receivablesDetailsForm-${useId()}`
```

MadMed677: Updated https://github.com/team-monite/monite-sdk/commit/eab3d83aca9da2d41e5cc17430d92f03da2378f0

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
MadMed677: Updated fc7432639ab94a4a829676230bb7df6fe0c94fd5

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
MadMed677: What do you mean? Permissions? All. All permissions are not supported on FE side for a new component

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
radist2s: You're quite right, the probability of having 2 `Monite-Form-receivablesDetailsForm` identifiers on the page is low. But why think about probabilities when it's better to use [the standard approach](https://react.dev/reference/react/useId#why-is-useid-better-than-an-incrementing-counter) that React offers

radist2s (approved):

PR #64, opened 2024-05-10
feat(DEV-10786): Receivables tables extra sorting

radist2s:


nekedos (approved):

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
```diff
@@ -163,12 +172,16 @@ const InvoicesTableBase = ({ onRowClick }: InvoicesTableProps) => {
               },
             },
             {
-              field: 'counterpart_id',
+              field: 'counterpart_name',
               headerName: t(i18n)`Customer`,
-              sortable: false,
+              sortable: receivableCursorFieldsList.includes(
```
MadMed677, line 177: Shouldn't we provide `onChangeSort` with the value as we do for the `PayablesTable` and the other ones?

MadMed677 (approved):

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
radist2s: I would remove [those callbacks](https://github.com/team-monite/monite-sdk/blob/b6dffd2d3cd2786d800b0e8f81135cb253b958c9/packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx#L63-L77) altogether unless sufficient arguments can be found in defense of the need for them. @nekedos, what do you think about that?

MadMed677: Unfortunately, it's not possible to know if our customers use it or not. But I am not the best but still a solution for it. We could add `Sentry` and send warnings if they do.
So if after 1-2 months there won't be any requests we could remove them and be sure that nobody uses it and it might be removed with no regrets.

What do you think?

radist2s: I think the less we give outside, the easier it will be for us to maintain. I would be wary of adding such a callback just to test a hypothesis. It would be better to wait for feedback from customers.

nekedos: Let's add `onChangeSort` later.

radist2s: @MadMed677, are you ok with that?

MadMed677: Sorry for such a delay. Forgot about that PR.
I mentioned that to remove callbacks which you mentioned [here](https://github.com/team-monite/monite-sdk/pull/64#discussion_r1599849536) it would be better to use Sentry to verify that nobody uses them. I think it would be better than just removing them without any proof.

I'm fine to not add `onChangeSort` here

@radist2s you may close this thread if you wish

PR #67, opened 2024-05-15
feat(DEV-10563): Improve test stability

MadMed677:
General overview of what has been changed:
1. Remove the custom `delay` and use the `msw` `delay` function. The main difference is that the custom delay has a timeout of 0ms. But the `msw` delay has a timeout of 5ms. It leads to the problem that some of our tests rely on immediate response and they have been fixed accordingly.
2. Introduce the `environmentSensitiveTimeout` function which should be used to provide timeouts only if it's not tested. Otherwise, our tests are slow and flacky.
3. Check buttons before we click on them. Sometimes we `click` on the button but they might be `disabled` because some BE requests haven't been completed.

Basically, that's all what have been done at that PR


packages/sdk-react/src/components/approvalPolicies/ApprovalPoliciesTable/ApprovalPoliciesTable.test.tsx
```diff
@@ -48,6 +48,8 @@ describe('ApprovalPoliciesTable', () => {
     test('should have "Created by user" trigger on the page', async () => {
       renderWithClient(<ApprovalPoliciesTable />);
 
+      await waitUntilTableIsLoaded();
```
radist2s, line 51: As I remember, @nekedos requested to avoid this pattern with the `waitUntilTableIsLoaded()`, and use instead something like `screen.findByRole(...)` on the target element.

radist2s (approved):

radist2s (approved):
💋

packages/sdk-react/src/components/approvalPolicies/ApprovalPoliciesTable/ApprovalPoliciesTable.test.tsx
MadMed677: We shouldn't avoid it :) `findByRole` waits for 1 second until something should show. But our test should represent what the user sees. So at the beginning the user sees `Spinner` so we have to wait until spinner disappears and then we have to wait until some async button shows.
The best option is to describe the same thing that our user sees and `waitUntilTableIsLoaded` is the best option (in my opinion) to represent it

radist2s: The problem is that there could be 2 spinners on the page, and I wouldn't want to get stuck with that logic.

nekedos: I have just commented `await waitUntilTableIsLoaded();` and this test starts working faster.

With `waitUntilTableIsLoaded`
Screenshot 2024-05-17 at 13 42 44

Without `waitUntilTableIsLoaded`
Screenshot 2024-05-17 at 13 43 06

@Artem-Monite, Can you check it?

MadMed677: > The problem is that there could be 2 spinners on the page, and I wouldn't want to get stuck with that logic.
I've already changed it and there is no problem. We just have to call `waitUntilTableIsLoaded` 2 times (example: https://github.com/team-monite/monite-sdk/pull/67/files#diff-d479f54e0f0c2151feab34fc82ea8082faf4aaa807bf70519563e79321bb5432R87)

MadMed677: > I have just commented await waitUntilTableIsLoaded(); and this test starts working faster.

I've checked it and this is my result (I've passed the test 5 times to get the average value which should show more correct data instead of 1 pass)


---
#### With `waitUntilTableIsLoaded`
- 490ms
- 599ms
- 496ms
- 501ms
- 499ms

**Average value: 517ms**

#### Without `waitUntilTableIsLoaded`
- 488ms
- 476ms
- 505ms
- 480ms
- 501ms

**Average value: 490ms**
---

Differences based on 5 time passes is 27ms. It's quite close to waitFor documentation of an interval of 50ms.
Based on [waitFor documentation](https://testing-library.com/docs/dom-testing-library/api-async#waitfor) the interval of checking is `50ms`.

So we wait for 1 tick more because we have 2 async actions instead of 1. In that specific case which doesn't consume a lot of time will work just `findBy` but we can't avoid `waitUntilTableIsLoaded` because it waits until the spinner disappears and only then will asynchronously waits when the text `Created by user` appears. Sometimes only `findBy` may fail because for 1 second MSW / spinner / text won't appear but when we have more granular checks it makes the code less error-prone but more stable.

nekedos: @Artem-Monite According to your logic, if we had three spinners, would we have to call `await waitUntilTableIsLoaded()` three times?

radist2s: I wouldn't get so worked up about performance in this case, that's not the problem.

MadMed677: > According to your logic, if we had three spinners, would we have to call await waitUntilTableIsLoaded() three times?

Yes. It's strict forward. The same as if we have to click on a button 3 times. We can't just set a timeout for 5ms and wait when something will appear. It won't give any information when something goes wrong.


> we are updating the interface and somewhere now an extra spinner is displayed (this has happened before). In this case, we will have to update all tests where the problem with two spinners will occur.

I think we have to find that problem and find out how to deal with it. Right now I found 1 case when we have 2 spinners and it's inside super-components.
The first spinner appears in the head of the header (see GIF)
The second spinner appears **after** the first is gone (see GIF)



So the code reacts on the first spinner and then we have to wait until the second spinner is gone. Maybe to make it more transparent we could handle not only spinners but also add some indicators to it. Like `progressbar` with some class or smth like that.

MadMed677: Also, we could react on the availability `Create` button. Right now I see it's available when the first spinner is gone. It's also a solution. Can't say it's better or worse

nekedos: @MadMed677 

> Yes. It's strict forward. The same as if we have to click on a button 3 times. We can't just set a timeout for 5ms and wait when something will appear. It won't give any information when something goes wrong.

So, if we have 100 rows in the Grid and each row has a spinner, do we need to call `waitUntilTableIsLoaded()` 100 times?

MadMed677: Why? I mean what's the point of that?
For me, if we have even 2 spinners to show something this is some UX problem. Show 100 spinners for the customer is quite strange behavior.

The idea of `waitUntilTableIsLoaded` is hidden inside the naming :) We wait until the table is loaded and then we can manipulate the data. If we have a spinner when we load as an example counterpart name or we show `skeleton` we should use `findBy` in that case because it doesn't matter what we show, we really wait until the data is available.

But we should use `waitUntilTableIsLoaded` because it blocks the whole UI and it may take time, but then some buttons / data also way load, and then we have to use `findBy` to wait when the button / data will be available but only after blocking spinner is hidden

nekedos: > For me, if we have even 2 spinners to show something this is some UX problem. Show 100 spinners for the customer is quite strange behavior.

We can’t guarantee that the component will have two or fewer spinners, but your solution requires us to know this 

> But we should use waitUntilTableIsLoaded

If you want to check how many spinners exist in the component, add a test that checks the number of spinners instead of hiding this check under the magic of the `waitUntilTableIsLoaded.

So, I can't agree this solution is good enough, sorry.

Can I ask you to replace `waitUntilTableIsLoaded` with `findBy*` or find a better solution?

MadMed677: > We can’t guarantee that the component will have two or fewer spinners, but your solution requires us to know this

Of course! When we use construction `waitFor -> expect(button).not.toBeDisabled()` we know about interface. We know that our UX disables the button and it should be available after some time. So the idea of waiting when the spinner disappears is the same. We know about UI we know that our UI is not available until the spinner goes away.

Look at this from another angle, we don't have a spinner we have a `Skeleton`. When the skeleton disappears some blocks might be available after that.

> Can I ask you to replace waitUntilTableIsLoaded with findBy*

I can't because some tests will fail. And they will because as [I mentioned before ](https://github.com/team-monite/monite-sdk/pull/67#discussion_r1606726665) default timeout is `1 second`. And:
1. We can't use `findBy` for sync buttons because it may describe the wrong action. As an example when the spinner goes away, the button is already here and we can just say `getBy*`. There is no async stuff after a table is loaded but if we wrap everything into `findBy` tests become slower and they won't represent the truth. They will say that every item on the page is asynchronous but it isn't true. But when we wait until the page is loaded (and we do that by waiting when the spinner goes away) everything then could be treated as sync (which is true)
2. In some scenarios when we have a lot of waiting for default timeout won't be enough. So we have to increase timeouts to 2 or 3 seconds. And the problem is that we are not waiting for some button for 3 seconds. We wait when spinner 1 goes away, then when spinner 2 goes away and only then do we wait when the async button appears. To make the test more reliable and to describe **what really happens** we have to describe what we are really waiting for. And right now we wait when the spinner to go away. This is the truth, isn't it? :) But when the spinner disappears we really in an async way wait when the button appears in that case U're right. We have to use `findBy` but only when other states are complete.


I don't know if it's clear but if it's not feel free to reach me out in DM I really hope I could explain in better.

nekedos (approved):
It is not ideal solution. Let's improve it later.

PR #70, opened 2024-05-20
style(DEV-10393): align pagination arrows center

saladNights:
fixed the pagination arrow's alignment
Screenshot 2024-05-20 at 13 14 43


MadMed677 (approved):

.changeset/five-balloons-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+align pagination arrows center
```
nekedos, line 5: Fix it, please, according to the rules:

### Changeset messages (changelog)

1. **Content:** Use present simple for entries.
2. **TASK-ID:** End each entry with the task identifier (e.g., *(DEV-1234)*).
3. **Examples:**
    - `fix(DEV-3456): add social network authentication feature`

packages/sdk-react/src/ui/table/TablePagination.tsx
```diff
@@ -148,39 +141,41 @@ export const TablePagination = <T,>({
...
-          <StyledSelect
-            {...slotProps?.pageSizeSelect}
-            aria-label={t(i18n)`Rows per page`}
-            MenuProps={{
-              ...slotProps?.pageSizeSelect?.MenuProps,
-              container: root,
+        {hasPageSizeSelect && (
+          <Box
+            sx={{
+              position: 'absolute',
```
radist2s, line 147: `position: absolute` could work, but `<StyledSelect />` could be configured using the theme. Therefore it is important not to break the layout.

Here is the overrided `<StyledSelect />` example:
image

I would suggest only adding an extra blank cell to be used for centering.
Probably, usage of `<FormControl size="small"/>` also not needed

radist2s (changes requested):
Thanks more for keeping things under control!

.changeset/five-balloons-fail.md
saladNights: fixed

packages/sdk-react/src/ui/table/TablePagination.tsx
saladNights: fixed

radist2s (approved):

PR #71, opened 2024-05-20
feat(DEV-8990): Improve InvoiceDetails screen

MadMed677:
## Description
Handle some problems:
1. Show the correct error window for PDF documents with the `refresh` button support
2. Don't show the Submit invoice section if the user can't `Issue` an invoice
3. Handle BE permissions and don't show buttons if there are any restrictions ([task](https://monite.atlassian.net/browse/DEV-10818))

Screenshot 2024-05-17 at 17 45 56
ar-pdf-error

### DIFF
--- a/.changeset/modern-cooks-pay.md
+++ b/.changeset/modern-cooks-pay.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+feat(DEV-8990): improve InvoiceDetails screen.
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
@@ -202,8 +202,6 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
     );
   }
 
-  console.log('--- pdfError: ', pdfError);
-
   return (
     <>
       <InvoiceDeleteModal
@@ -348,11 +346,14 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
                   i18n
                 )`You don't have permission to issue this document. Please, contact your system administrator for details.`}</Alert>
               ) : (
-                <SubmitInvoice
-                  deliveryMethod={deliveryMethod}
-                  onDeliveryMethodChanged={setDeliveryMethod}
-                  disabled={loading}
-                />
+                (buttons.isIssueButtonVisible ||
+                  buttons.isComposeEmailButtonVisible) && (
+                  <SubmitInvoice
+                    deliveryMethod={deliveryMethod}
+                    onDeliveryMethodChanged={setDeliveryMethod}
+                    disabled={loading}
+                  />
+                )
               )}
               <Overview invoice={receivable} />
             </Stack>
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
@@ -27,6 +27,7 @@ import KeyboardArrowRightIcon from '@mui/icons-material/KeyboardArrowRight';
 import EmailIcon from '@mui/icons-material/MailOutline';
 import MoreVertIcon from '@mui/icons-material/MoreVert';
 import {
+  Alert,
   Box,
   Button,
   DialogContent,
@@ -123,10 +124,21 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
     props.id
   );
 
+  const { data: isUpdateAllowed } = useIsActionAllowed({
+    method: 'receivable',
+    action: ActionEnum.UPDATE,
+    entityUserId: receivable?.entity_user_id,
+  });
+
   /** Is the deleting modal opened? */
   const [deleteModalOpened, setDeleteModalOpened] = useState<boolean>(false);
 
-  const { data: pdf, isLoading: isPdfLoading } = usePDFReceivableById(props.id);
+  const {
+    data: pdf,
+    isLoading: isPdfLoading,
+    error: pdfError,
+    refetch: refetchPdf,
+  } = usePDFReceivableById(props.id);
 
   const handleIssueAndSend = useCallback(() => {
     setPresentation(InvoiceDetailsPresentation.Email);
@@ -189,6 +201,8 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
       />
     );
   }
+
+  console.log('--- pdfError: ', pdfError);
 
   return (
     <>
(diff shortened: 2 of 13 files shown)


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -143,6 +143,18 @@ const EmailInvoiceDetailsBase = ({
     ]
   );
 
+  const isDisabled = useMemo(() => {
```
radist2s, line 146: @MadMed677, I'm already embarrassed to mention this every time, but such operations don't need to be memoized. Using `useMemo(...)` here only adds semantic load, and actually only slows down the code. I won't point out other places where you refactored and left the unreasonable use of `useMemo(...)` as it is, but it would be nice if you fixed that too.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
```diff
@@ -293,25 +322,39 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
               sx={{
                 display: 'flex',
                 flexDirection: 'column',
-                height: '100%',
+                // height: '100%',
```
radist2s, line 325: Probably, it would be better to remove it; or uncomment it?

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
```diff
@@ -138,14 +188,14 @@ const FileViewerComponent = ({
...
+
   const renderFile = () => {
     if (isPdf)
       return (
-        <Document
-          file={url}
-          onLoadSuccess={onDocumentLoadSuccess}
-          onLoadError={console.error} //TODO add error view when design will be ready
-        >
+        <Document file={url} onLoadSuccess={onDocumentLoadSuccess}>
```
radist2s, line 198: Shouldn't we display some error message using `onLoadError` props if the document has some problems?

.changeset/modern-cooks-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+feat(DEV-8990): improve InvoiceDetails screen. Support backend permissions for different buttons. Don't show some blocks if there is no data for it
```
radist2s, line 5: May I as you to split Changesets to separate MD files? The final result will look much better as three different features.

radist2s (approved):
🤘

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
MadMed677: U're right, forgot to remove it. Thank you

.changeset/modern-cooks-pay.md
MadMed677: What do you mean? Could you provide an example? I don't get it :)

MadMed677: You mean call another `yarn changeset add` and based on that create another `.md` file with different message, is it correct?

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
MadMed677: This is a quite strange library :) It doesn't call any `onLoadError`, `onSourceError`, or any other callbacks if there is no `url` has been specified ([based on react-pdf documentation](https://github.com/wojtekmaj/react-pdf?tab=readme-ov-file#props)).
So the only thing is to provide `noData` but this is when we have no file by provided `url` so I've checked `url` before sending any request to remove any PDF controls that we have right now.
So that's how it looks like:
Screenshot 2024-05-24 at 15 40 45

So I could provide `noData` "just for safety" if our BE returns 200 with `url` but without any PDF by that link (I don't think that it's a real case but I have no problem with that check either)

Commit: [226b339](https://github.com/team-monite/monite-sdk/pull/71/commits/226b339db677840fd716a6395a4a35062e204cf7)

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
MadMed677: [cb8dce4](https://github.com/team-monite/monite-sdk/pull/71/commits/cb8dce4e1771b6cf5894bb5ab233ca84bd488359)

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
MadMed677: I do not agree with that point though. But to avoid any holy wars I will change it.
But I'll change my mind if you provide any proof in our code that this `useMemo` slows down the code and provides no benefits.

Commit: [7fda03b](https://github.com/team-monite/monite-sdk/pull/71/commits/7fda03b35d2791e88649cfac65cb1b2bbd8bc005)

nekedos: **From official documentation:**

[Should you add useMemo everywhere? ](https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere)

Optimizing with useMemo  is only valuable in a few cases:

- The calculation you’re putting in useMemo is noticeably slow, and its dependencies rarely change.
- You pass it as a prop to a component wrapped in [memo.](https://react.dev/reference/react/memo) You want to skip re-rendering if the value hasn’t changed. Memoization lets your component re-render only when dependencies aren’t the same.
- The value you’re passing is later used as a dependency of some Hook. For example, maybe another useMemo calculation value depends on it. Or maybe you are depending on this value from [useEffect.](https://react.dev/reference/react/useEffect)

There is no benefit to wrapping a calculation in useMemo in other cases. There is no significant harm to doing that either, so some teams choose to not think about individual cases, and memoize as much as possible. The downside of this approach is that code becomes less readable. Also, not all memoization is effective: a single value that’s “always new” is enough to break memoization for an entire component.

.changeset/modern-cooks-pay.md
radist2s: Sorry for the late answer. Yep, just for the beauty in the Changelog, it would be cool to see different bullets on different features.

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
radist2s: Thanks for the clarification!

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
radist2s: @MadMed677, tanks, Solved.

.changeset/modern-cooks-pay.md
MadMed677: Splitted them into 2 different changesets and 2 different commits:
- 1644a535912136a32c9616f8b027bdc9e527702c
- https://github.com/team-monite/monite-sdk/pull/71/commits/cb9c44c1acbd8a75b52fa7d578eea055085ce29d

radist2s (approved):

PR #72, opened 2024-05-20
feat(DEV-10391): add `<InvoicesTable/>` Quick Actions 🎬

radist2s:
This PR introduces a new **Actions Menu** to the `InvoicesTable` component, providing users with a convenient way to interact with _individual invoices directly from the table_.

**Key Improvements:**

* **Enhanced User Experience:** Users can now access common actions like **view**, **edit**, and **delete** for each invoice through a dedicated menu (assuming `onRowActionClick` is defined)
  * If no `rowActions` prop is not provided, the table will display a set of default actions
* **Flexible Configuration:** The `rowActions` prop allows customization of the available actions based on the invoice status. You can define a set of actions for each `ReceivablesStatusEnum`.

#### Usage Examples

**1. Custom Row Actions:**
```tsx
<InvoicesTable 
    onRowActionClick={customOnRowActionClick} 
    rowActions={{[ReceivablesStatusEnum.DRAFT]: ['view']}} 
/>
```

**2. Default Actions:**
```tsx
<InvoicesTable onRowActionClick={customOnRowActionClick} />
```

This improvement significantly enhances the usability of the `InvoicesTable` component, making it easier and more intuitive for users to manage their invoices.

#### Screeshot
image


packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+      () =>
+        useInvoiceRowActionMenuCell({
+          onRowActionClick: jest.fn(),
+        }),
+      {
+        wrapper: createRenderWithClient(),
+      }
+    );
+
+    await waitFor(() => checkPermissionQueriesLoaded(testQueryClient));
```
MadMed677, line 32: Let's remove `checkPermissions*` check from the code. It's an outdated approach. Please use `await waitUntilTableIsLoaded()`

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
```diff
@@ -0,0 +1,252 @@
...
+    context: 'InvoicesTableRowActionMenu',
+  }),
+});
+
+interface InvoicesTableRowActionMenuItem {
+  label: string;
+  action: InvoicesTableRowAction;
+}
+
+export type InvoicesTableRowAction =
```
MadMed677, line 209: It would be nice to provide TSDoc for each of options. As I see this type will be provided to the customer and I think it's better to provide TSDoc for each of actions (especially custom ones) to get the customer know what & how it will affect them

packages/sdk-react/src/core/queries/usePermissions.ts
```diff
@@ -59,35 +59,54 @@ export function useIsActionAllowed({
...
+export function isActionAllowed({
+  action,
+  actions,
+  entityUserId,
+  entityUserIdFromAuthToken,
+}: {
+  action: PayableOperator['action'] | CommonOperator['action'];
+  actions: Array<ActionSchema | PayableActionSchema> | undefined;
+  entityUserId?: string;
+  entityUserIdFromAuthToken: string | undefined;
```
MadMed677, line 96: Let's move it into a type and cover it by TSDoc. What each field means. I think it would be helpful.

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
```diff
@@ -36,15 +36,56 @@ import { GridSortDirection } from '@mui/x-data-grid/models/gridSortModel';
...
    * The event handler for a row click.
    *
    * @param id - The identifier of the clicked row, a string.
    */
   onRowClick?: (id: string) => void;
-};
+
+  /**
+   * The event handler for an invoice row action.
+   * If not provided, the action menu will not be displayed.
```
MadMed677, line 55: As I understand we have 2 options:
1. User doesn't provide `onRowActionClick` & `rowActions`
2. User provides `onRowActionClick` & `rowActions`

Because if the user provides `onRowActionClick` we can't show row actions. And vice-versa if the customer provides `rowActions` but not `onRowActionClick` we can't show UI with no callbacks.

So my idea is to have a united type, as an example:

```ts
interface InvoicesTableInternalProps {
  onRowClick?: (id: string) => void;
}

interface InvoicesTableRowActionsProps extends InvoicesTableInternalProps {
  // Make it required!
  onRowActionClick: ...
  
  rowActions?: ...
}

type InvoicesTableProps = InvoicesTableInternalProps | InvoicesTableRowActionsProps
```

In that case, if the customer would like to provide `rowActions` they have to specify the `onRowActionClick` callback. If they don't so in that case everything will be the same.

[We already use it for](https://github.com/team-monite/monite-sdk/blob/29d69da72605a506ac29437592d606829158bf01/packages/sdk-react/src/components/receivables/InvoiceDetails/InvoiceDetails.types.ts#L100) `InvoiceDetails` component as an example when we specify `create` or `update` scenarios

So WDYT?

packages/sdk-react/src/components/receivables/InvoicesTable/InvoiceActionMenu/InvoiceActionMenu.stories.tsx
```diff
@@ -0,0 +1,102 @@
+import { ExtendThemeProvider } from '@/utils/ExtendThemeProvider';
```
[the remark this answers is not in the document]
radist2s: Sorry I didn't warn you, but looking through the commits wasn't the best idea. It was a tough PR with a lot of bad decisions along the way. Better to only watch the final result. `InvoiceActionMenu.stories.tsx` does not exist anymore.

packages/sdk-react/src/components/receivables/InvoicesTable/InvoiceActionMenu/InvoiceActionMenu.tsx
```diff
@@ -0,0 +1,390 @@
...
+ *       styleOverrides: {
+ *         root: {
+ *           outline: 'solid red 1px',
+ *         },
+ *       },
+ *     },
+ *   },
+ * });
+ */
+export const InvoiceActionMenu = memo((inProps: InvoiceActionMenuProps) => {
```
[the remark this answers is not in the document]
radist2s: `InvoiceActionMenu` not exists anymore

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.test.tsx
radist2s: No, I don't think so. In my opinion - `waitUntilTableIsLoaded(...)` is a semantic garbage. The idea is that a table spinner just isn't enough, since actions depend on the permissions load state. Unfortunately, I haven't found any way to otherwise ensure that permissions have been fully loaded for _this case_.

packages/sdk-react/src/core/queries/usePermissions.ts
radist2s: TSDoc already exists, external type not needed.

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
radist2s: Thank you for the great idea! Refactored [in another way](https://github.com/team-monite/monite-sdk/blob/abe0d7ebbdb4dc4435fef87fd6b55368e34aa4d6/packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx#L18-L54).

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
radist2s: Added, thanx!

packages/sdk-react/src/components/receivables/InvoicesTable/InvoiceActionMenu/InvoiceActionMenu.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+        }}
+        invoice={{
+          id: '1',
+          status: ReceivablesStatusEnum.DRAFT,
+        }}
+        onClick={onClick}
+      />
+    );
+
+    await waitFor(() => checkPermissionQueriesLoaded(testQueryClient));
```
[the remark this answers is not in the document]
radist2s: Outdated, but I have got the idea, thank you. Will try the approach in the next time.

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.test.tsx
radist2s: Removed, since the component logic was completely changed.

nekedos (approved):

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
```diff
@@ -0,0 +1,308 @@
...
+    inActions?.[invoice.status] ?? DEFAULT_ACTION_LIST[invoice.status];
+
+  const allowedActions = filterInvoiceActionMenuAllowedItems(
+    receivableActionSchema,
+    actions,
+    invoice,
+    userIdFromAuthToken
+  );
+
+  const allMenuItems = getInvoiceActionMenuItemLabels(i18n);
```
MadMed677, line 120: Maybe we should memoize the result from this function? As I can see we call it on each click but we don't need to create a new object every time when the user clicks on an item. We could safe the returned value (especially because `i18n` never changes) and reuse the result without generating a new object

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
```diff
@@ -0,0 +1,308 @@
...
+  invoice: InvoiceResponsePayload;
+  actions?: Partial<InvoicesTableRowActionSchema>;
+  receivableActionSchema: ActionSchema[] | undefined;
+  userIdFromAuthToken: string | undefined;
+  i18n: I18n;
+}): InvoicesTableRowActionMenuItem[] => {
+  const actions =
+    inActions?.[invoice.status] ?? DEFAULT_ACTION_LIST[invoice.status];
+
+  const allowedActions = filterInvoiceActionMenuAllowedItems(
```
MadMed677, line 113: The same as the previous question. Maybe it's better to memoize the response from this function? Otherwise, we have to create a new object and make 13 loops (inside `menuItemsPermissionMap` we call `isAllowedInvoiceAction` which makes a loop for all actions and then calls `filter`)

I think memoization should work much better to save memory and avoid extra loops.

WDYT?

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+      expect(
+        filterInvoiceActionMenuAllowedItems(
+          fullPermissions,
+          menuItems,
+          {
+            ...receivableListFixture.invoice[0],
+            status: ReceivablesStatusEnum.DRAFT,
+            entity_user_id: 'user_id_1',
+          },
+          'user_id_1'
```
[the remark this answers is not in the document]
radist2s: If the test passes with these variables, it works. The mocker user id not needed here.

packages/sdk-react/src/components/receivables/ReceivablesTable/ReceivablesTable.tsx
```diff
@@ -10,12 +10,12 @@ import { CreditNotesTable } from '../CreditNotesTable';
 import { InvoicesTable } from '../InvoicesTable';
 import { QuotesTable } from '../QuotesTable';
 
-interface ReceivablesTableUncontrolledProps {
+interface IReceiveablesTableUncontrolledProps {
```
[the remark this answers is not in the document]
radist2s: Thank you, it was the wrong rebase resolution with the earlier commits.

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
```diff
@@ -105,6 +116,12 @@ const InvoicesTableBase = ({ onRowClick }: InvoicesTableProps) => {
     setCurrentPaginationToken(null);
   };
 
+  const invoiceActionCell = useInvoiceRowActionMenuCell({
+    rowActions: 'rowActions' in restProps && restProps.rowActions,
+    onRowActionClick:
```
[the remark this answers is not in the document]
radist2s: That's great to find, need to take note. Unfortunately, this is the default behavior of `<GridActionsCellItem/>` (no backdrop) and I wouldn't want to change these components to a custom menu like in Counterparts.

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.stories.tsx
```diff
@@ -13,6 +13,7 @@ type Story = StoryObj<typeof InvoicesTable>;
 export const FullPermissions: Story = {
```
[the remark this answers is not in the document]
radist2s: Thank you for the idea, added an extra example.

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
radist2s: I don't think that it is so needed:

```
1. getInvoiceActionMenuItems: 0.0068359375 ms
2. getInvoiceActionMenuItems: 0.006103515625 ms
3. getInvoiceActionMenuItems: 0.005859375 ms
4. getInvoiceActionMenuItems: 0.005126953125 ms
5. getInvoiceActionMenuItems: 0.006103515625 ms
6. getInvoiceActionMenuItems: 0.005859375 ms
7. getInvoiceActionMenuItems: 0.006103515625 ms
8. getInvoiceActionMenuItems: 0.0048828125 ms
9. getInvoiceActionMenuItems: 0.006103515625 ms
10. getInvoiceActionMenuItems: 0.005126953125 ms
```

JS is just fast enough with the operations on simple data structures.

packages/sdk-react/src/components/receivables/InvoicesTable/useInvoiceRowActionMenuCell.tsx
radist2s: The same, JS is just fast on loops. It's a kinda built-in feature. Otherwise, why would we have so many ways to iterate?

```
1. filterInvoiceActionMenuAllowedItems: 0.0029296875 ms
2. filterInvoiceActionMenuAllowedItems: 0.006103515625 ms
3. filterInvoiceActionMenuAllowedItems: 0.003173828125 ms
4. filterInvoiceActionMenuAllowedItems: 0.0029296875 ms
5. filterInvoiceActionMenuAllowedItems: 0.001953125 ms
6. filterInvoiceActionMenuAllowedItems: 0.0029296875 ms
7. filterInvoiceActionMenuAllowedItems: 0.001953125 ms
8. filterInvoiceActionMenuAllowedItems: 0.0029296875 ms
9. filterInvoiceActionMenuAllowedItems: 0.001953125 ms
10. filterInvoiceActionMenuAllowedItems: 0.0048828125 ms
11. filterInvoiceActionMenuAllowedItems: 0.001953125 ms
```

packages/sdk-react/src/components/receivables/ReceivablesTable/ReceivablesTable.tsx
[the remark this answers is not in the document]
MadMed677: NP

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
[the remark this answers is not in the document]
MadMed677: Okay, got it

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.stories.tsx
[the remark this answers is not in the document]
MadMed677: Danke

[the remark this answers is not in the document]
MadMed677: Danke

PR #81, opened 2024-05-27
feat(DEV-10943): add Invoices Table "Due date" filter ◍

radist2s:
Implemented a _"Due Date"_ filter for the Invoices Table to enhance user experience by enabling efficient searching and filtering of invoices based on due dates.

#### Changes

1.  **Invoices Table Update**:
    
    -   Added the _"Due Date"_ filter to the _Invoices Table_, allowing users to filter invoices by due dates.
2.  **Column Update in Invoices Table**:
    
    -   Replaced the `fulfillment_date` column with the `due_date` column in the `<InvoicesTable />` component.
3.  **API Query Parameter**:
    
    -   Integrated the `due_date__lte` parameter to support filtering options.
#### Usage

Users can now filter invoices due before or on a specific date using the `due_date__lte` parameter.

image

### DIFF
--- a/packages/sdk-react/src/components/receivables/Filters/Filters.tsx
+++ b/packages/sdk-react/src/components/receivables/Filters/Filters.tsx
@@ -1,10 +1,7 @@
 import React from 'react';
 
 import { counterpartsToSelect } from '@/components/payables/PayableDetails/PayableDetailsForm/helpers';
-import {
-  FilterTypes,
-  FilterValue,
-} from '@/components/receivables/ReceivablesTable/types';
+import { type ReceivablesFilterHandler } from '@/components/receivables/Filters/useReceivablesFilters';
 import { useRootElements } from '@/core/context/RootElementsProvider';
 import { useCounterpartList } from '@/core/queries';
 import { SearchField } from '@/ui/SearchField';
@@ -26,7 +23,7 @@ import {
 } from '../consts';
 
 type Props = {
-  onChangeFilter: (field: keyof FilterTypes, value: FilterValue) => void;
+  onChangeFilter: ReceivablesFilterHandler;
 };
 
 export const Filters = ({ onChangeFilter }: Props) => {
@@ -55,7 +52,9 @@ export const Filters = ({ onChangeFilter }: Props) => {
             onChange={(selected) => {
               onChangeFilter(
                 FILTER_TYPE_STATUS,
-                selected && selected.target.value
+                selected.target.value === 'all'
+                  ? null
+                  : (selected.target.value as ReceivablesStatusEnum)
               );
             }}
           >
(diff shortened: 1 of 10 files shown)


packages/sdk-react/src/components/receivables/Filters/Filters.tsx
```diff
@@ -18,118 +16,152 @@ import {
...
+                  selected.target.value === 'all'
+                    ? null
+                    : (selected.target.value as ReceivablesStatusEnum)
+                );
+              }}
+            >
+              {[
+                { label: t(i18n)`All statuses`, value: 'all' },
+                ...Object.values(ReceivablesStatusEnum).map((status) => ({
+                  label: (() => {
```
MadMed677, line 71: What do you think of moving this logic into a function constant? To not re-create an object on each update.
We could create a function constant by `key: ReceivablesStatusEnum` and a value `t(i18n)`

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
```diff
@@ -220,7 +238,7 @@ const InvoicesTableBase = ({ onRowClick }: InvoicesTableProps) => {
               flex: 0.5,
             },
             {
-              field: 'fulfillment_date',
```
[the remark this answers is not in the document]
MadMed677: Yo, no we can't. `fulfillment_date` is a date when the invoice has been created. But `due_date` is a date for the customer when they have to pay for this invoice.

As an example of a salary invoice. You create an invoice to Monite 29th of May and set `due_date` as `01.06` for Monite.

MadMed677 (changes requested):

packages/sdk-react/src/components/receivables/InvoicesTable/InvoicesTable.tsx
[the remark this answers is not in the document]
MadMed677: But this is a matter of what to show. In my humble opinion `due_date` is much more important than `fulfillment_date` so I think we could change it.
Because we don't have a product owner right now I think we could change it with no hesitation

MadMed677 (approved):

packages/sdk-react/src/components/receivables/Filters/Filters.tsx
radist2s: I don't see a problem with creating this object and iterating each time. It's not a 1000-line list.

MadMed677: But it's anyway worse than creating a function 1 time and not re-creating it on each render. If we will create a new function in each component it will affect performance. I see no point in not moving it into a separate function to avoid any memory costs when it's not needed

saladNights (approved):

packages/sdk-react/src/components/receivables/Filters/Filters.tsx
radist2s: You're talking about premature optimization. Memoization is much more expensive than iterating an object that is created and quickly deleted. Also, all this is wonderfully optimized by all known JS engines.

It's worth it to optimize the components that really slow things down. Also, React Compiler will address the need for optimization on its own in the future. And this is not some distant future.

PR #82, opened 2024-05-28
feat(proposal): improve `demo-data-generator` 🧬

radist2s:
#### Changes

1.  **CLI Commands**:
    
    -   Added `run` and `entity` commands to the `demo-data-generator` CLI.
3.  **Documentation**:
    
    -   Updated the `demo-data-generator` CLI documentation to reflect the new commands and their usage.
5.  **Data Generation**:
    -   Increased the number of items generated to ensure a more comprehensive dataset for testing.
    -   Added more payment terms to the demo data generation process to provide a broader range of test scenarios.


examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
```diff
@@ -46,6 +48,115 @@ const commandWithEntityOptions = () =>
...
+      .description('Create a new Entity')
+      .option(
+        '--generate-demo-data',
+        'Generate demo data for the new entity',
+        true
+      )
+      .option('--client-id <ID>', 'The client_id to use', String)
+      .option('--client-secret <ID>', 'The client_secret to use', String)
+      .action(async (args) => {
+        const token = await fetchTokenCLI(args);
```
MadMed677, line 81: What will happen if `await` throws an error? Does the `commander` library handle it and print errors in the console?

Shouldn't we wrap it into `try / catch` and in `catch` print an error into the console & to Clerk UI?

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
```diff
@@ -46,6 +48,115 @@ const commandWithEntityOptions = () =>
...
+
+          if (id)
+            console.log(
+              chalk.greenBright(
+                `👤 Created new entity_user_id: "${id}" with role: "${role}"`
+              )
+            );
+        }
+
+        if (args.generateDemoData) await generateDemoData(newEntityId, token);
```
MadMed677, line 151: I'd prefer to use `{}` for all if / else statements for readability reasons. But it's a matter of code style and we don't have any rules so if you don't want to do it I have no permissions to force you

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
```diff
@@ -397,6 +492,28 @@ program
       })
   );
 
+const generateDemoData = async (entity_id: string, token: AccessToken) => {
```
MadMed677, line 495: To be honest I don't like the naming because it doesn't provide enough information about what is going to do that function.

Based on what is happening in that function I may suggest the following options:
- Rename it into `initialization` (or `init` if you prefer) - because we check `entity_id`, open mqqt messenger to write UI messages & generating an entity. So from my point of view, it looks like initialization and should be called before all API methods
- Add `TSDoc` with a description (less preferable for me) and describe what it going on here. Other engineers may avoid reading the code and will understand it based on documentation

But I prefer `initialization`.

What do you think?

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/generate-entity.ts
```diff
@@ -69,13 +69,14 @@ export const generateEntity = async (
     const measureUnitsService = new MeasureUnitsService(
       serviceConstructorProps
     );
-    const measureUnits = await measureUnitsService.create();
+    await measureUnitsService.create();
```
MadMed677, line 72: What is the point of creating measure units but after that fetches them again?
`create` will create `N` measure units to use them. But as I can see you create them but then fetch them again `.getAll()`. Is there any reason for that?

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/generate-entity.ts
```diff
@@ -99,7 +100,7 @@ export const generateEntity = async (
 
     const receivablesService = new ReceivablesService(serviceConstructorProps);
 
-    /** Create 15'ing Receivables with a type `invoice` */
+    /** Create 20'ing Receivables with a type `invoice` */
```
MadMed677, line 103: [not required] Maybe I've over tsdoc it :) I think this TSDoc may be removed

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
radist2s: If we find a sufficient reason to do so and enshrine it in a code-styled guide... In all other cases, it's a matter of taste.

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
radist2s: I changed the name to `generateEntityDemoData(...)` since this is not initialization after all (it does not create Entity Users and Roles). I also added a TSDoc.

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/generate-entity.ts
radist2s: If an Entity already has measure units with the same names, the creation will not go any further. Sometimes it is convenient to run Invoice generation multiple times.

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/generate-entity.ts
radist2s: ☑︎

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/generate-entity.ts
MadMed677: Hm, maybe then we should run `measureUnitsService.create()` only once to avoid any duplicates and provide that structure to `generate-entity` and reuse created values. Or as you did: create it once but inside `generate-entity` just call `getAll()` and take a few to add it to invoices

Anyway it's not so important

examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/bin.ts
MadMed677: > it's a matter of taste

yep, yep, as I mentioned. No forcing to change it

MadMed677 (approved):

PR #84, opened 2024-05-28
feat(DEV-10563): Remove `environmentSensitiveTimeout` from the code to fix sdk-demo

MadMed677:


.changeset/bright-items-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+feat(DEV-10563): safely take `process` if it's not available
```
nekedos, line 5: It would be better to remove the changeset because it wasn't published.

Use `yarn changeset add --empty` instead.

radist2s (approved):

.changeset/bright-items-walk.md
MadMed677: Just removed the commit

nekedos (approved):

PR #87, opened 2024-05-29
feat(DEV-10025): Show loader when PDF url is absent

MadMed677:
## Description
When Receivable has been changed (we issue it or edit it) the BE will return `file_url: null`. And if it happens we have to re-fetch it until we will have it.

#### That how it looks like:
ar-pdf

Also, there is an option when we still can't load PDF after 4 iterations. In that case, we show our current error with an ability to refresh PDF manually
ar-pdf-2


radist2s (changes requested):
Hey, this is great stuff! I have a [suggestion with improvements](https://github.com/team-monite/monite-sdk/pull/89).

MadMed677:
> Hey, this is great stuff! I have a [suggestion with improvements](https://github.com/team-monite/monite-sdk/pull/89).

So based on what I see on PR the main idea is to move [this code](https://github.com/team-monite/monite-sdk/pull/87/files#diff-116b42126f21bcbe47e20eabcd1e66bb82c7582d1f9b2164774a8205508a4d00R329-R342) in a separate component?

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -163,6 +163,11 @@ export const useIssueReceivableById = () => {
       queryClient.setQueryData(receivablesQueryKeys.detail(id), receivable);
       invalidate();
 
+      /** We have to invalidate a PDF query to get the latest PDF */
+      queryClient.invalidateQueries({
```
saladNights, line 167: optional: if you have time, you could add invalidation to `useEntityCache`, as it is implemented in `useEntityListCache`, and re-use it here.

saladNights (approved):

packages/sdk-react/src/core/queries/useReceivables.ts
MadMed677: Done: [f549c7e](https://github.com/team-monite/monite-sdk/pull/87/commits/f549c7eb750cab6c458851bda38c1096bda182ae)

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -357,19 +362,42 @@ export const usePDFReceivableById = (
...
     queryKey: receivablesQueryKeys.pdf(receivableId),
+    queryFn: async () => {
+      const response = await monite.api.receivable.getPdfLink(receivableId);
+
+      if (!response.file_url) {
+        /**
+         * We have to flush current query data to show the error
+         *  but not previous data
+         */
+        queryClient.setQueryData(receivablesQueryKeys.pdf(receivableId), null);
```
radist2s: Can refactor the custom result handling from `queryFn`? I don't see the need to use such a complicated pattern, it seems it would be easier to use `refetchInterval`

MadMed677: Can't say that `refetchInterval` is nicer but I tried but the problem is that `refetchInterval` will call BE again and again even if BE fails many times.

So in my current approach (at least for me) everything is straightforward. Query function should describe what should be done to fetch the data. So in our case if there is no `file_url` we have to fetch the data again but no more than 4 times just to avoid any BE overload.

So in `retry` we could tackle it. If we constantly have 4 errors in a row we will throw an error inside `QueryFn` which is `File URL is not provided` and this error will be provided to public API inside `error`.

But can't implement so for the `refetchInterval` because there is no way to provide an error to the customer.
[Also I see that `retry` is designed for this logic](https://tanstack.com/query/latest/docs/framework/react/guides/query-retries#retry-delay) to describe should we `retry` or not based on condition. And if we throw an error in `queryFn` we could manage it

radist2s: Your approach complicates future Qraft integration. Of course, a custom `queryFn` could be specified for its hooks as well. But why, when you can do the same with the approach I suggested?

MadMed677: Okay, but how would you fix the problem of stopping iterating after N times and providing an error to the customer? As I have in the current implementation? To stop pulling BE repeatedly if they provide no PDF and show an error to the client at the same time.

radist2s: Don't you think that's a pretty rare case? Maybe, on the contrary, it's worth trying time after time to get the link if it's empty? Empty link !== API error.

MadMed677: > Don't you think that's a pretty rare case?

I think it's important to cover all possible scenarios we never know when PDF will be generated. Maybe it will never will be generated because the customer provided some wrong data and our BE may never return the data. So answering your question: I think it's a rare case but it doesn't mean that we shouldn't handle it.

> it's worth trying time after time to get the link if it's empty

Yep, in the current approach, we try 4 times and if it still fails we show an error and provide a button (see in `.gif` file) when the user may manually click and re-fetch it.

---

So I see a few problems with `refetchInterval`:
1. If BE responds with an error still call BE over and over again no matter what
2. If BE responds with an empty `file_url` we also call BE over and over again no matter what

As I can see by [Tanstack documentation](https://tanstack.com/query/latest/docs/framework/react/guides/query-retries#retry-delay) `retry` is designed specifically for `queryFn` errors because it provides `failureCount` and `error` and based on that we could decide should we `retry` or not. But `refetch` is a general terminology responsible for refetching data not retrying failed ones.

But if it's so important for you to avoid `retry` and write it inside `refetchInterval` would you mind helping me with these 2 problems that I mentioned above? How do we handle errors and avoid any BE requests?

radist2s: It seems we just have different perspectives on how requests from APIs should be handled.

packages/sdk-react/src/core/queries/useReceivables.ts
radist2s: In my opinion `useReceivablePdfListCache` is overcomplicating.

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -34,6 +34,16 @@ const receivablesQueryKeys = {
...
 };
 
+const useReceivablePdfListCache = () => {
+  const queryClient = useQueryClient();
+
+  return {
+    invalidate: (receivableId: string) => {
+      queryClient.setQueryData(receivablesQueryKeys.pdf(receivableId), null);
+    },
+  };
```
radist2s, line 44: I think the previous implementation was better because it would be easier to fix in the future.

MadMed677: Moved it back [289e24e](https://github.com/team-monite/monite-sdk/pull/87/commits/289e24e46beae79fbab15a4d0d0b167deed7b20b)

packages/sdk-react/src/core/queries/useReceivables.ts
MadMed677: Moved it back: [289e24e](https://github.com/team-monite/monite-sdk/pull/87/commits/289e24e46beae79fbab15a4d0d0b167deed7b20b)

radist2s: Invalidation keeps the previous query in the cache. When we update the Invoice, probably, the previous result for the PDF must be not just invalidated, but reset, to show spinner immediately instead of the old PDF.

MadMed677: So then I think it's better to have a separate function to handle those query client requests
[c1e18c8](https://github.com/team-monite/monite-sdk/pull/87/commits/c1e18c8e7e7434340e198d3b482c35ef15e7d9e6)

packages/sdk-react/src/core/queries/useReceivables.ts
MadMed677: [c1e18c8](https://github.com/team-monite/monite-sdk/pull/87/commits/c1e18c8e7e7434340e198d3b482c35ef15e7d9e6)

radist2s (approved):

PR #89, opened 2024-05-29
feat(proposal): changes to DEV-10025 (Invoice PDF link refreshing) ♨️

radist2s:
1.  Refactor: Componentize Invoice File Viewer
2.  Refactor: Simplify `usePDFReceivableById(...)` hook logic
3.  Feature: Clear PDF Viewer After Invoice Update

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { memo, useCallback, useState } from 'react';
 
 import { useDialog } from '@/components';
 import { EditInvoiceDetails } from '@/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails';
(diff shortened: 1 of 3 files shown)


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
```diff
@@ -373,3 +344,47 @@ const ExistingInvoiceDetailsBase = (props: ExistingReceivableDetailsProps) => {
     </>
   );
 };
+
+const InvoiceFileViewer = memo(({ invoiceId }: { invoiceId: string }) => {
+  const { data, isLoading, error, refetch } = usePDFReceivableById(invoiceId);
+  const { i18n } = useLingui();
+
+  if (error)
```
MadMed677, line 352: `FileViewer` already handles error if `url` is not provided. So I think we don't need to handle an alert for it because `FileViewer` will show an error and an ability to `Refresh` request if user clicks on the button

radist2s: I'm not sure if `FileViewer` displays the error that occurs during the query. The query may crash, for example, why not tell the user about it?

MadMed677: If `url` is not provided FileViewer will show an error. Or if PDF is not provided, or if PDF is not valid, etc
It has been made here: https://github.com/team-monite/monite-sdk/pull/71

PR #94, opened 2024-05-31
feat(DEV-8691): Add frontend validation for products with provided `small amount`

MadMed677:
## Description
Handle `smallest_amount` if what the user provided `amount` is less than `smallest_amount` we throw an error and forbid to submission of the form

ar-smallest-amount


radist2s (approved):
Thank you!

PR #95, opened 2024-06-02
feat(DEV-11582): deploy preview

radist2s:


Caddyfile.pull-preview
```diff
@@ -0,0 +1,25 @@
+{
+    auto_https disable_redirects
```
vkuprin, line 2: Is there a specific reason for disabling HTTPS redirects?

Caddyfile.pull-preview
```diff
@@ -0,0 +1,25 @@
...
+}
+
+{$PULLPREVIEW_URL} {
+    reverse_proxy web:3000
+
+    handle_path /__clerk/* {
+        uri strip_prefix /__clerk
+        reverse_proxy https://frontend-api.clerk.dev {
```
vkuprin, line 10: Maybe, consider parameterizing the URL to make it easier to change in the future. The same as is done for https://{$PULLPREVIEW_PUBLIC_DNS}/__clerk

PR #110, opened 2024-06-11
chore(DEV-11178): Update ESLint Rule mui-require-container-property

vkuprin:
* Added logic in `mui-require-container-property.ts` to specifically handle cases where `useMenuButton(...)` is used to spread props into MUI components, ensuring the `container` property is included if missing.
* Expanded test coverage in `mui-require-container-property.test.ts` to include scenarios with `useMenuButton(...)` hook, particularly checking for the presence of the `container` property in spread operations.
* Modified ESLint configurations in `.eslintrc.json` within `sdk-react` package to switch the rule from "off" to "error", enforcing strict compliance across the codebase.
* Adjusted various components to directly include or correct the `container` property in their spread attributes to adhere to the updated rule, reducing potential rendering issues in popper-based components.

### DIFF
--- a/packages/eslint-plugin/src/tests/rules/mui-require-container-property.test.ts
+++ b/packages/eslint-plugin/src/tests/rules/mui-require-container-property.test.ts
@@ -43,6 +43,11 @@ ruleTester.run<string, readonly unknown[]>(
                <DatePicker />`,
       },
       {
+        code: `import { Menu } from '@mui/material';
+           const config = { menuProps: { container: 'root' } };
+           <Menu {...config.menuProps} />`,
+      },
+      {
         code: `import { DatePicker } from '@mui/x-date-pickers';
                <DatePicker
                  slotProps={{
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin/src/tests/rules/mui-require-container-property.test.ts
```diff
@@ -153,6 +168,18 @@ ruleTester.run<string, readonly unknown[]>(
           },
         ],
       },
+      {
+        code: `import { Select } from '@mui/material';
+               const props = {};
+               <Select {...props} />`,
```
radist2s, line 174: Why should we extend test cases with `<Select {...props} />`? Maybe I'm wrong, isn't it `useMenuButton(...)` used only for the `<Menu {...menuProps} />`?

vkuprin: You're correct, i'm just testing edge cases on pipelines

radist2s: Ok, man, I won't review the draft. I'm just excited.

nekedos (approved):

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/ExistingInvoiceDetails.tsx
```diff
@@ -63,6 +63,7 @@ const StyledMenu = styled((props: MenuProps) => {
         vertical: 'top',
         horizontal: 'right',
       }}
+      container={props.container}
```
radist2s, line 66: @vkuprin, here you must just pass `useRootElements().root` instead of `props.container`.
There is no chance that `StyledMenu` will receive the `container` property, so, just force it as:
```
container={useRootElements().root}
```

radist2s: Nope, I'm wrong, it will receive property from `<StyledMenu {...menuProps}>`.
Anyway, I would add the default property as a fallback. The `menuProps` will override it for sure too.

PR #111, opened 2024-06-12
fix(DEV-10644): prevent theme selector overlap when menu exceeds viewport height

edcs:
image


saladNights (changes requested):

radist2s (approved):
Thank you!

saladNights (approved):
all good

PR #123, opened 2024-06-18
feat(DEV-9911): add GitLab submodule sync with the `dev` GitHub branch instead of `main` 🏗️

radist2s:
This PR introduces changes to make the `dev` branch the default branch and updates the workflow for Changesets and synchronization processes. The main points are as follows:

1. **Default `dev` Branch**:
   - The `dev` branch is now set as the default branch.
   - All changes to the new default `dev` branch must go through a Pull Request process to ensure code quality and consistency.

2. **Changesets `baseBranch` Update**:
   - The `baseBranch` for Changesets is set to `origin/dev` in the `dev` branch to ensure correct calculation of Changesets during development.
   - A validation job is implemented to ensure the `baseBranch` is correctly set. If not updated to `origin/main` before merging `dev` branch changes into `main`, the job will trigger an alert.

3. **Synchronization with GitLab**:
   - Changes from the `dev` branch are synchronized with the GitLab `dev` branch, from which deployment to Dev stand will occur.
   - Changes from the `main` branch will be synchronized with the GitLab default branch.

4. **Sentry Secret Variables**:
   - Added usage of Sentry secret variables to fix sourcemap deployment issues.


.changeset/three-points-collect.md
vkuprin, line 1: I think we need a comment explaining why the base branch is being changed to origin/dev

.changeset/three-points-collect.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
vkuprin, line 2: I think we need a comment explaining why the base branch is being changed to origin/dev

nekedos (approved):

vkuprin (approved):

.changeset/three-points-collect.md
nekedos: But it is about our internal development process. We don't publish it to `npm`.

PR #126, opened 2024-06-20
feat(DEV-9445): style main menu with monite theme

edcs:
image


examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const selected = pathname === href?.replace(window.location.origin, '');
+
+  const buttonProps =
+    href === undefined
+      ? { component: 'button', onClick }
+      : { component: 'a', href };
+
+  return (
+    <ListItem className="navigation-list-item">
+      <ListItemButton
```
radist2s, line 36: We must use `component={Link}` from the `import Link from 'next/link';` to be compatible with Next.js routing.

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const pathname = usePathname();
+  const selected = pathname === href?.replace(window.location.origin, '');
+
+  const buttonProps =
+    href === undefined
+      ? { component: 'button', onClick }
+      : { component: 'a', href };
+
+  return (
+    <ListItem className="navigation-list-item">
```
radist2s, line 35: We prefer to to not add an extra `className` until we need it.

packages/sdk-themes/monite-theme.d.ts
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+
+import '@mui/material/styles';
+import '@mui/x-data-grid/themeAugmentation';
+
+declare module '@mui/material/styles' {
+  interface TypographyVariants {
+    label1: React.CSSProperties;
+    label2: React.CSSProperties;
+    label3: React.CSSProperties;
```
radist2s, line 10: We must not add extra variants.
Our customers will be obliged to use such `augmentation` somehow in their applications, and they will certainly use our types for their MUI components as well.

I understand that your decision was dictated by design, but the discrepancies between design and architecture need to be addressed at the design stage - we are completely tied to the MUI types and had no plans to change anything in favor of design.

packages/sdk-themes/src/themes/monite.ts
```diff
@@ -6,12 +6,14 @@ import type {
 } from '@mui/material/styles/createPalette.js';
 import type { Theme, ThemeOptions } from '@mui/material/styles/createTheme.js';
 import type { TypographyOptions } from '@mui/material/styles/createTypography.js';
-import type {} from '@mui/x-data-grid/themeAugmentation';
+
+import { neutralsTransparent, primary } from '../colors/monite/index.js';
```
radist2s, line 10: If you don't mind, let's maintain `colors` in the same file with the theme. I'm not sure that we will share the palette.

examples/with-nextjs-and-clerk-auth/src/app/globals.css
```diff
@@ -2,4 +2,5 @@ body,
 html {
   margin: 0;
   padding: 0;
+  min-height: 100vh;
```
radist2s, line 5: I would prefer to not fix the layout (`min-height`), while it is possible. It's better to fix a wrapper.

radist2s:
By the way, please put "Draft" for PR if it's not already finished.

package.json
```diff
@@ -51,5 +51,8 @@
     "@types/react": "18.3.1",
     "@types/react-dom": "18.3.0"
   },
-  "packageManager": "yarn@3.5.0"
+  "packageManager": "yarn@3.5.0",
+  "dependencies": {
+    "@iconscout/react-unicons": "~2.0.2"
```
radist2s, line 56: I realize this was added by mistake, but let's be more attentive to detail when submitting work to a review.

examples/with-nextjs-and-clerk-auth/src/components/ThemeSelect/ThemeSelect.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { FC, SyntheticEvent, useRef, useState } from 'react';
+
+import {
+  UilSun,
```
radist2s, line 4: Haven't you checked what the is console showing us?

image

radist2s (changes requested):
Earlier, you posted this PR in Ready status and I later moved it to Draft on my own as I noticed serious flaws after checking it. Could you please clarify if this PR is still in the works? Would you consider the current status as final?

image

image

package.json
edcs: Please keep code reviews polite - this is not the way to make change suggestions.

examples/with-nextjs-and-clerk-auth/src/components/ThemeSelect/ThemeSelect.tsx
edcs: I am aware of this issue, it is from the icon library which I am required to use to fulfil the goals of this project.

examples/with-nextjs-and-clerk-auth/src/themes/material.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { Theme } from '@mui/material/styles/createTheme';
+
+const components: Components<Omit<Theme, 'components'>> = {
+  MuiList: {
+    styleOverrides: {
+      root: {
+        '&.navigation-list': {
+          '.MuiCollapse-root .MuiListItemButton-root': {
+            paddingLeft: 28,
+          },
```
radist2s, line 12: I'd rather you add styling directly in the component, rather than at the theme level. The MUI theme should be exactly the same as in `packages/sdk-themes`:

```ts
import type { ThemeOptions } from '@mui/material/styles/createTheme.js';
export const materialLight: ThemeOptions = { palette: { mode: 'light' } };
export const materialDark: ThemeOptions = { palette: { mode: 'dark' } };
```

examples/with-nextjs-and-clerk-auth/src/components/ThemeSelect/ThemeSelect.tsx
```diff
@@ -0,0 +1,140 @@
...
+        >
+          <Box component="span" display="flex" color="neutral.10">
+            {mode === 'dark' ? <IconMoon /> : <IconSun />}
+          </Box>
+          <Box component="span" display="flex" flexDirection="column" flex="1">
+            <Typography variant="body2" component="span" textAlign="left">
+              {themeName}
+            </Typography>
+            <Typography
+              className="theme-select__mode"
```
radist2s, line 76: As I have already advised, it is better to remove all `className` and use css-in-js features. We try not to rely on classes as they are harder to maintain in case of using ready-made component libraries.

examples/with-nextjs-and-clerk-auth/src/monite-theme.d.ts
```diff
@@ -0,0 +1,59 @@
...
+      '90': string;
+      '95': string;
+      '100': string;
+    };
+  }
+
+  interface TypographyVariants {
+    label1: React.CSSProperties;
+    label2: React.CSSProperties;
+    label3: React.CSSProperties;
```
radist2s, line 43: As I might have noticed, `label1` is not used anywhere for styling.
`label2` is used in three places, and in only two of the cases in different contexts.
`label3` is used 1 time.
Furthermore, I don't see a significant difference in styles in these variants, which contradicts the idea of style tokens.
I'd recommend focusing on tweaking the sentient options, and where standard options aren't enough - styling components where required.

examples/with-nextjs-and-clerk-auth/src/monite-theme.d.ts
```diff
@@ -0,0 +1,59 @@
...
+import '@mui/material/styles';
+import '@mui/x-data-grid/themeAugmentation';
+
+declare module '@mui/material/styles' {
+  interface TypeBackground {
+    default: string;
+    paper: string;
+    menu: string;
+    highlight: string;
+  }
```
radist2s, line 12: It doesn't make sense to me to add all these types just for the sole purpose of using it in a single place:
```ts
PaperProps={{ sx: { backgroundColor: 'background.menu' } }}
```

Furthermore, if you want to change the styles for all menus, why not do it? Why style menus in Demo one way and in SDK intentionally another way? It seems like the idea of Demo is to show that widgets can be styled any way you want.
I recommend focusing on customizing the theme MUI than creating new entities that require complex support.

examples/with-nextjs-and-clerk-auth/src/monite-theme.d.ts
```diff
@@ -0,0 +1,59 @@
...
+      '10': string;
+      '30': string;
+      '50': string;
+      '70': string;
+      '80': string;
+      '90': string;
+      '95': string;
+      '100': string;
+    };
+  }
```
radist2s, line 38: Most of the types here are for changes that are needed in one or a couple places, and they all involve menus, which in my opinion is over-complication.

radist2s (changes requested):

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const NavigationListItem = ({
+  endIcon,
+  children,
+  href,
+  icon,
+  onClick,
+}: NavigationLinkProps) => {
+  const pathname = usePathname();
+  const selected = pathname === href?.replace(window.location.origin, '');
```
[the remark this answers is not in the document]
radist2s: I was thinking about [a more convenient way](https://nextjs.org/docs/app/api-reference/functions/use-pathname) for the next.js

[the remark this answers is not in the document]
radist2s: @edcs, `'use client'` does not say that this component is executed only on the client. It says that the code of this component is delivered to the client, instead of raw html markup as in the case of Server Components.

<details><summary>console error log</summary>
<p>
image
</p>
</details> 

Please be more careful and check every now and then what is being dumped to the console.

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+  href,
+  icon,
+  onClick,
+}: NavigationLinkProps) => {
+  const pathname = usePathname();
+  const selected = pathname === href;
+
+  const buttonProps =
+    href === undefined
+      ? { component: 'span', onClick }
```
radist2s, line 38: We need to pass a custom component only in the case if we pass the `href`. In the same time, the link can also handle `onClick` callbacks.
```ts
const buttonProps =
    href !== undefined ? { component: Link, href } : undefined;
// ..
<ListItemButton selected={selected} {...buttonProps}>
```

examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
```diff
@@ -0,0 +1,165 @@
...
+import type { TypographyOptions } from '@mui/material/styles/createTypography';
+import {
+  paletteDark as basePaletteDark,
+  paletteLight as basePaletteLight,
+  defaultMoniteComponents,
+  defaultMoniteTypography,
+} from '@team-monite/sdk-themes';
+
+const paletteLight: PaletteOptions = {
+  ...basePaletteLight,
```
radist2s, line 16: By using destructuring, you break the standard theme gluing mechanism.
In the MUI, in order to extend an external theme with your own styles, you need to pass styles as the second argument to `createTheme(defaultTheme, {/* one more theme */})`.

Check out the [API](https://mui.com/material-ui/customization/theming/#api).

packages/sdk-themes/monite-theme.d.ts
radist2s: To summarize: at the meeting, we came to the understanding that additional options could be added to simplify future styling when new themes are added, but we only expand the options to those that are used.
Thanks for clarifying the path in making your decision.

examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
```diff
@@ -0,0 +1,165 @@
...
+      },
+    },
+  },
+  MuiButton: {
+    styleOverrides: {
+      root: {
+        '&.theme-select': {
+          borderRadius: 8,
+        },
+        '&.theme-select .theme-select__mode': {
```
radist2s, line 132: I think it's better to stick to the class naming style like the MUI, for example, we do the [same](https://github.com/team-monite/monite-sdk/blob/1c5be62c37a8916839a27956d35da336aa989765/packages/sdk-react/src/components/receivables/InvoiceStatusChip/InvoiceStatusChip.tsx#L88).
```
.ThemeSelect .ThemeSelect-modeLabel
```

examples/with-nextjs-and-clerk-auth/src/components/ThemeSelect/ThemeSelect.tsx
```diff
@@ -0,0 +1,122 @@
...
+            onClick={() => handleVariantChange({ variant: 'monite', mode })}
+          >
+            <Typography variant="body1">{t(i18n)`Monite`}</Typography>
+          </ThemeSelectMenuItem>
+          <Divider />
+          <ThemeSelectMenuItem>
+            <Typography variant="label2">{t(i18n)`Language`}</Typography>
+          </ThemeSelectMenuItem>
+          <ThemeSelectMenuItem checked>
+            <Typography variant="body1">{t(i18n)`English`}</Typography>
```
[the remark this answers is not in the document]
radist2s: As we have no more langs, for now, it's ok.

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListCollapse.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { PropsWithChildren, ReactNode, useState } from 'react';
+
+import { Collapse, List } from '@mui/material';
+
+import { NavigationListItem } from '@/components/NavigationMenu/NavigationListItem';
+import { IconAngleDown, IconAngleUp } from '@/icons';
+
+type NavigationListCollapseProps = PropsWithChildren<{
```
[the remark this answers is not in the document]
radist2s: It's more productive from a DX perspective, and in my opinion `PropsWithChildren` is harder for the reader to understand, which is more important than a nanosecond on the CPU.
If you pay attention, we don't use `PropsWithChildren` in any place in our project.

If it's a matter of principle for you, you can keep it. I only give a recommendation to match the general style of the monorepository codebase.

examples/with-nextjs-and-clerk-auth/src/monite-theme.d.ts
radist2s: Just the opposite. I would like SDK to be able to reuse all styles from Demo as much as possible. Ideally - it would not be clear where Demo ends and SDK begins.
As for `<Drawer/>` - you can style it globally, through the theme. Especially since it is not used in the SDK (and if it was - it would be cool to get similar colors for it).

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationList.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  return (
+    <>
+      <List className="navigation-list" disablePadding>
+        <NavigationListItem href="/" icon={<IconApps />}>
+          {t(i18n)`Dashboard`}
+        </NavigationListItem>
+        <NavigationListItem href="#" icon={<IconUsdCircle />}>
+          {t(i18n)`Purchases`}
+        </NavigationListItem>
```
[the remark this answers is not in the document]
radist2s: Solution: leave the items without action for now.

examples/with-nextjs-and-clerk-auth/src/components/NavigationMenu/NavigationListItem.tsx
radist2s: Sorry for the previous incorrect example, you missed `onClick`:
```tsx
<ListItemButton selected={selected} onClick={onClick} {...buttonProps}>
```

examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
```diff
@@ -0,0 +1,171 @@
...
+      paper: {
+        border: 'none',
+        borderRadius: 16,
+        width: 240,
+      },
+    },
+  },
+};
+
+export const moniteLight = createTheme(
```
radist2s, line 157: When you perform:

```ts
export const moniteLight = createTheme(...)
```

it automatically calls `createTheme()` right at the moment of importing the file, which is not critical, but I think it's better to do it once in the place where you import a theme.

examples/with-nextjs-and-clerk-auth/src/components/ThemeSelect/ThemeSelect.tsx
radist2s: Clarified, let's keep the selector if it's needed.

examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
```diff
@@ -0,0 +1,171 @@
...
+  label3: {
+    color: paletteDark.neutral && paletteDark.neutral['80'],
+  },
+});
+
+const components: Components<Omit<Theme, 'components'>> = {
+  MuiDrawer: {
+    styleOverrides: {
+      root: {
+        '&.navigation-drawer .MuiPaper-root': {
```
radist2s, line 97: Maybe naming for the class such as `LayoutNavigationDrawer` would be a better fit. I would prefer a casing like in the MUI.

radist2s:
By the way, when you think a PR is ready for a review, move it to Ready. Last time I moved it to Draft because I thought you opened a PR in Ready by accident.

examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
radist2s: Yeah, sure, you can use `deepmerge` at this point to merge different themes. But I see no reason to call `createTheme()` as many times as we import themes. You can do this once: `<ThemeProvider theme={createTheme(theme)} />`. Also, you already call `createTheme()` in [`AppThemeProvider`](https://github.com/team-monite/monite-sdk/blob/e4d41e7e112b6dc96ead64e7d4a78a0531dab3db/examples/with-nextjs-and-clerk-auth/src/components/ThemeRegistry/AppThemeProvider.tsx#L110-L120).

radist2s (approved):

nekedos (approved):

PR #130, opened 2024-06-24
feat(DEV-11277): add new Monite Iframe App 🔲

radist2s:
This PR introduces a new Monite Iframe App and enhances existing functionality:

### Key Features

- **Iframe App Elements**: Implement core components for the Iframe application.
- **SDKDemoAPIProvider**: Introduce a new provider component and integrate Qraft in `SDKDemo` for improved state management.
- **Drop-In Components**:
  - Add support for the `disabled` property, allowing deferred component loading. This enables programmatic changes to attributes like `basename` or `entity-id` before the component renders.
  - Optimize bundle size for improved performance.

### Additional Improvements
- Update documentation to reflect new features and integration processes.
- Refactor and optimize existing code.


radist2s:
Make note that "Sync PR with GitLab" is failing because the GitLab branch contains specific changes and the sync tool will not overwrite them.

packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
```diff
@@ -0,0 +1,86 @@
...
+        document.removeEventListener('click', clicker);
+      };
+    },
+    [fetchToken]
+  );
+
+  return { ...slots, fetchToken };
+};
+
+function validateToken(token: unknown): token is {
```
nekedos, line 74: It is better to use the existing `AccessToken` type here.

packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
```diff
@@ -0,0 +1,86 @@
...
+      iframeCommunicator.connect();
+
+      return () => {
+        iframeCommunicator.disconnect();
+      };
+    },
+    [iframeCommunicator, fetchToken]
+  );
+
+  useEffect(
```
nekedos, line 56: I think it is possible to skip using `clicker` and use `fetchToken` directly.

```ts
useEffect(
  function subscribe() {
    document.addEventListener('click', fetchToken);

    return () => {
      document.removeEventListener('click', fetchToken);
    };
  },
  [fetchToken]
);
```

packages/sdk-drop-in/src/lib/fetchTokenDev.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createAPIClient } from '@monite/sdk-react';
+
+export async function fetchTokenDev() {
+  const { entity_user_id, client_id, client_secret, api_url } =
+    await getConfig();
+  const { api, requestFn } = createAPIClient();
+  return await api.auth.postAuthToken(
+    {
+      baseUrl: `${api_url}/v1`,
+      parameters: {},
```
vkuprin, line 11: Are we forced to pass an empty object? Can we just remove it as it's seems redundant

packages/sdk-drop-in/src/lib/MoniteIframeAppCommunicator.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (MoniteIframeAppCommunicator.allowedSlots.includes(data.type))
+      this.listeners.get(data.type)?.(data.payload);
+
+    function assertData(
+      data: unknown
+    ): asserts data is { type: string; payload?: ListenerPayload } {
+      if (typeof data !== 'object' || data === null) {
+        throw new Error(`Invalid message data: ${data}`);
+      }
+      if (typeof (data as { type: unknown }).type !== 'string') {
```
vkuprin, line 79: Should we handle a cases when other extentions are enabled and iterupt this payload? 

Screenshot 2024-07-02 at 17 51 57

packages/sdk-drop-in/src/lib/MoniteIframeAppCommunicator.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    if (!this.targetWindow) throw new Error('Target window is not set');
+
+    window.addEventListener('message', this.onWindowMessage);
+
+    if (!this.isHost) {
+      this.connectionToHostRetryInterval = setInterval(() => {
+        if (!this.targetWindow) throw new Error('Target window is not set');
+        this.connect();
+      }, 200) as unknown as number;
```
vkuprin, line 43: Can we move this 200 to top of the class as a `const `to follow some clean code principles and give it describtion that is `const intervalConnectionTime = 200` ?

packages/sdk-drop-in/src/lib/fetchTokenDev.ts
radist2s: Unfortunately, because of Qraft's type-system `parameters` are required. I have ideas on simplifying such cases, but this requires architecture changes and will be in the next version.

packages/sdk-drop-in/src/lib/MoniteIframeAppCommunicator.ts
radist2s: Yep, we have to think about that too. We will find a way on how to do this. Probably, it's possible to check the origin of the event source, but we have to compare the origin with the customer domain, which is not easy (but doable).

radist2s: Btw, released to your comment: I refactored `console.log` usage to avoid displaying sensitive data.

packages/sdk-drop-in/src/lib/MoniteIframeAppCommunicator.ts
radist2s: I don't think this will add any value to the code.
- We are not going to change the value in the runtime
- value used just once
- value isn’t needed to have an understanding of what the class is doing

packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
radist2s: Thank you, removed this garbage. This is why I asked for the PR.

packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
radist2s: Replaced in each place in the `sdk-drop-in`.

PR #132, opened 2024-06-25
feat(DEV-11282): Playwright Setup for Chrome (iframe)

vkuprin:
- Set up Playwright for Chrome.
- Implement a test to verify that the sales page is rendered.
- Verify navigation functionality within the iframe.
- Add a case to fetch the token and render the iframe.


packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,45 @@
+import { test } from '@playwright/test';
+
+const getMockConfig = () => ({
+  client_id: '2afafd88-7b8c-4ed2-b0f0-652c0db600b7',
+  client_secret: 'ae6b5ece-51a1-4fcf-97d2-d10bb0de0612',
+  entity_user_id: '0771f748-2e98-400e-adb4-20ebefd77f5f',
+  grant_type: 'entity_user',
+});
```
radist2s: We should not expose secrets, even development. Usually, we use MSW (as in `@monite/sdk-react` package). But this package is not yet ready to implement MSW tests probably. So, for the local environment let's use `.env.local` file with predefined env variables. For the CI stage, we will use GitHub secrets.

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    await page.route(`https://api.dev.monite.com/${authTokenPath}`, (route) => {
+      route.fulfill({
+        contentType: 'application/json',
+        body: JSON.stringify(tokenResponse),
+      });
+    });
+  });
+
+  test('should see the sales and navigate inside the iframe', async ({
```
radist2s, line 34: As part of iframe performance tests, do you think it's necessary to add user scenarios? I thought it was enough to test the basic functionality.

turbo.json
```diff
@@ -96,8 +96,21 @@
     },
     "codegen": {
       "cache": true,
-      "outputs": ["src/api/**/*.ts", "!src/api/client.ts", "!src/api/client.test.ts"],
+      "outputs": [
+        "src/api/**/*.ts",
+        "!src/api/client.ts",
+        "!src/api/client.test.ts"
+      ],
```
radist2s, line 103: If you don't mind, it's better not to include changes/refactoring not relevant to the task in the PR. Even a formatting change already requires the reviewer's attention.

radist2s (commented):
Look's great, thank you!

packages/sdk-drop-in/tests/iframe.test.ts
vkuprin: My idea is to check if routing inside the iframe actually works and some basic actions can be performed. Additionally, by including user scenarios, we can ensure comprehensive coverage and validate performance under realistic conditions

turbo.json
vkuprin: Sure, let's cover CI stuff in a next PR then

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  type WidgetTest = {
+    path: string;
+    name: string;
+  };
+
+  const widgetTests: WidgetTest[] = [
+    { path: routingPaths.payables, name: 'Payables' },
+    { path: routingPaths.counterparts, name: 'Counterparts' },
+    { path: routingPaths.products, name: 'Products' },
+  ];
```
radist2s, line 107: In such simple case all you need is:

```
  const widgetTests = [
    { path: routingPaths.payables, name: 'Payables' },
    { path: routingPaths.counterparts, name: 'Counterparts' },
    { path: routingPaths.products, name: 'Products' },
  ] as const;
```

But if you really need it to be typed, there's no shame to inline:
```
const widgetTests: {
    path: string;
    name: string;
  }[] = [
    { path: routingPaths.payables, name: 'Payables' },
    { path: routingPaths.counterparts, name: 'Counterparts' },
    { path: routingPaths.products, name: 'Products' },
  ];
```

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  widgetTests.forEach(({ path, name }) => {
+    test(`should see the ${name.toLowerCase()} tab`, async ({ page }) => {
+      await page.goto(`${consumerPage}${path}`);
+      const iframe = page.frameLocator('iframe');
+      await iframe.getByRole('heading', { name }).click();
```
radist2s, line 113: We don't particularly have to care about roaming in the Consumer App. But we should care that the buttons in the menu lead to the right Super Components inside iframe:

```tsx
await page.getByRole('button', { name }).click(); // click on the menu menu item
await iframe.getByRole('heading', { name }).toBeVisible(); // check if heading is rendered
```

turbo.json
```diff
@@ -95,6 +95,15 @@
     "clean": {
       "cache": false
     },
+    "e2e": {
+      "cache": false,
+      "persistent": true,
```
radist2s, line 100: As far as I understand, this task is not persistent, since it can terminate after the end of the job (unlike `dev` for example)

.github/workflows/monorepo-main-suite.yaml
```diff
@@ -69,9 +73,27 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Monorepo
+        uses: team-monite/setup-yarn-project-action@b4b4161e23724bfce41a9a66cb46be87cdd39e36
+
+      - name: Install Playwright Browsers
+        run: yarn dlx playwright install --with-deps
+
+      - name: Check dir content inside
+        run: ls -la packages/sdk-drop-in
```
radist2s, line 92: I assume this code is for debugging purposes.

radist2s (changes requested):

packages/sdk-drop-in/tests/iframe.test.ts
vkuprin: Thank you for sharing these insights. I was unaware that we already have test coverage within the SDK-React. In that case, it is indeed redundant

.github/workflows/monorepo-main-suite.yaml
```diff
@@ -69,9 +73,27 @@ jobs:
...
+    needs: build
+    name: E2E Playwright Testing
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Monorepo
+        uses: team-monite/setup-yarn-project-action@b4b4161e23724bfce41a9a66cb46be87cdd39e36
+
+      - name: Install Playwright Browsers
+        run: yarn dlx playwright install --with-deps
```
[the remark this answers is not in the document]
vkuprin: Maybe, i'm missing something, but it doesn't worked that way, just receiving timeout issues:

Screenshot 2024-07-01 at 10 44 37

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  test('test the theme switcher and tabs', async ({ page }) => {
+    await page.getByRole('button', { name: 'Material UI' }).click();
+    await page.getByText('Theme').click();
+    await page.getByRole('button', { name: 'Material UI' }).click();
+    await page.getByRole('menuitem', { name: 'Monite' }).click();
+    await page.getByRole('button', { name: 'Monite' }).click();
+    await page.getByLabel('Dark Mode').check();
+    await page.locator('.MuiBackdrop-root').click();
```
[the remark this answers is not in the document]
vkuprin: Added new changes for color check

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      (el) => getComputedStyle(el).backgroundColor
+    );
+
+    expect(bgColor).toBe('rgb(18, 18, 18)');
+  });
+
+  test('test the Roles button under Settings', async ({ page }) => {
+    await page.getByRole('button', { name: 'Settings' }).click();
+    await expect(page.getByRole('button', { name: 'Roles' })).toBeVisible();
+    await page.getByRole('button', { name: 'Roles' }).click();
```
radist2s, line 63: Could you also add a check to make sure the component(iframe) is actually loaded?
```ts
await expect(iframe.getByRole('heading', { name })).toBeVisible();
```
It would be great to test items working in this way too, the same as you did with `should see the ${name.toLowerCase()} tab`

packages/sdk-drop-in/playwright.config.ts
```diff
@@ -0,0 +1,35 @@
...
+  workers: process.env.CI ? 1 : undefined,
+  reporter: [['html', { open: 'never' }]],
+  use: {
+    actionTimeout: 0,
+    baseURL: 'http://localhost:5273',
+    trace: 'on-first-retry',
+    headless: true,
+  },
+  webServer: {
+    command: 'yarn dev --port 5273',
```
radist2s, line 21: May I ask you to add a comment that `yarn build && yarn preview` is not used here as routing support is needed, which is implemented for `dev` mode?

packages/sdk-drop-in/public/config.json
radist2s, line 1: This file is not needed.

radist2s (changes requested):

packages/sdk-drop-in/playwright.config.ts
vkuprin: Done, comment added

packages/sdk-drop-in/public/config.json
vkuprin: This is necessary; otherwise, the tests will fail. We need to mock data and intercept it. You can verify this by running tests without this file—Playwright will be unable to intercept the token. Essentially, this file mocks the data and allows interception with the real one from the CI environment. If a request is made to /config.json and the file is not in the public folder, it will return a 404 status code, preventing interception and body modification.

As an alternative, I will add logic to dynamically create that file during these specific tests

vkuprin: Added extra logic to add on fly config.json, please contact me for extra qustions, if you have alternative solution

.changeset/serious-suns-pay.md
radist2s, line 1: This changeset should be empty, as it does not affect the release.

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  const frame = await iframeElement?.contentFrame();
+
+  if (!frame) {
+    await page.reload({ waitUntil: 'domcontentloaded', timeout: 60000 });
+  }
+
+  await frame?.waitForURL(new RegExp('.*/receivables.*', 'i'), {
+    timeout: 60000,
+  });
+});
```
radist2s, line 71: I'd like to ask you not to add test logic to `test.beforeEach(...)`. `test.beforeEach(...)` is the place for mocks, but not for implementing the business logic of the test.

In your case, if I move the test "__test the theme switcher__" to another file, it will fail with a timeout. Although it requires a simple `await page.goto(...)`, without which the test becomes incomprehensible.

Sorry, I didn't pay attention to this last time.

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+test('test the theme switcher', async ({ page }) => {
+  const iframe = page.frameLocator('iframe');
```
radist2s, line 74: Could you clarify each case with the specified `await page.goto(...)` call?
Such a test would be more portable and easier to read for developers/QAs unfamiliar with Playwright.

packages/sdk-drop-in/tests/iframe.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  await iframe.locator('body').waitFor({ state: 'visible', timeout: 10_000 });
+
+  await page.getByRole('button', { name: 'Material UI' }).click();
+  await page.getByText('Theme').click();
+  await page.getByRole('button', { name: 'Material UI' }).click();
+  await page.getByRole('menuitem', { name: 'Monite' }).click();
+  await page.getByRole('button', { name: 'Monite' }).click();
+  await page.getByLabel('Dark Mode').check();
+  await page.locator('.MuiBackdrop-root').click();
```
radist2s, line 83: Is this possible to just click on the button again to close the menu?
```ts
await page.getByRole('button', { name: 'Material UI' }).click();
```

radist2s (approved):
Thank you! This is the most important brick in improving the quality of our product.

nekedos (approved):

PR #133, opened 2024-06-30
fix(DEV-11506): prevent incorrect rendering of images in `<FileViewer />` 😶‍🌫️

radist2s:
## Problem
When an image was selected instead of a PDF for an invoice, the `PDFObject.embed()` function was still being called, even though the PDF container wasn't rendered. This caused the entire page content to be replaced with the image inside an iframe, as the `ref` wasn't properly set.

## Solution
We've refactored the `FileViewer` component to correctly handle different file types, particularly distinguishing between PDFs and images.

## Changes

- Added a check to prevent `PDFObject.embed()` calls for non-PDF files
- Implemented separate rendering logic for PDFs and images:
  - For PDFs: Use `PDFFileViewer` component
  - For images: Render a simple `` tag with appropriate styling

## Impact
- Prevents unintended page content replacement when viewing non-PDF files
- Simplifies the codebase by removing redundant conditional checks

## Testing
Please test the following scenarios:
1. Uploading and viewing PDF files for invoices
2. Uploading and viewing image files for invoices
3. Switching between PDF invoice and image file invoice to ensure correct rendering
4. Checking that the page layout remains intact when viewing non-PDF files

### DIFF
--- a/packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
+++ b/packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
@@ -67,9 +67,26 @@ export const FileViewer = ({
   name,
   onReloadCallback,
 }: FileViewerProps) => {
+  if (!url) return <ErrorComponent onError={onReloadCallback} />;
+
+  if (mimetype === 'application/pdf') return <PdfFileViewer url={url} />;
+
+  return (
+    <img
+      src={url}
+      alt={name}
+      loading="lazy"
+      style={{ width: '100%', objectFit: 'contain' }}
+    />
+  );
+};
+
+const PdfFileViewer = ({ url }: { url: string }) => {
   const pdfRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
+    if (!pdfRef.current) return;
+
     PDFObject.embed(url, pdfRef.current, {
       fallbackLink: true,
       pdfOpenParams: {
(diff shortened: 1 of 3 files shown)


vkuprin (approved):

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
```diff
@@ -67,9 +67,26 @@ export const FileViewer = ({
   name,
   onReloadCallback,
 }: FileViewerProps) => {
+  if (!url) return <ErrorComponent onError={onReloadCallback} />;
+
+  if (mimetype === 'application/pdf') return <PdfFileViewer url={url} />;
```
vkuprin, line 72: I'm wondering, if we should handle a case for mimetypes === 'text/plain' or microsoft office files with application/vnd. ... (docx)  (xlsx) or (pptx)

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
```diff
@@ -67,9 +67,26 @@ export const FileViewer = ({
...
+
+  return (
+    <img
```
vkuprin, line 75: Does it makes sense to have a test coverage for this for a different mime types as well? To check that we render what we actually want and pass

As for insipiration to discuss, but this for only one case:

```ts
  test('renders image tag when mime type is an image', () => {
    const { rerender } = render(<FileViewer url="test.jpg" mimetype="image/jpeg" name="test" onReloadCallback={errorCallback} />);
    
    expect(screen.getByRole('img')).toHaveAttribute('src', 'test.jpg');

    rerender(<FileViewer url="test.png" mimetype="image/png" name="test" onReloadCallback={errorCallback} />);
    expect(screen.getByRole('img')).toHaveAttribute('src', 'test.png');
  });
```

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
radist2s: As I know, our API supports images or PDFs only.

packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
radist2s: @vkuprin, thank you, [add](https://github.com/team-monite/monite-sdk/pull/133/commits/9ce56a93218dff581e836c620fb529afa1a2d236). Do you have ideas on how we could test iframe embedding?

vkuprin: I'm thinking of utilising our playwright setup

nekedos (approved):

PR #154, opened 2024-07-10
feat(DEV-11730): migrate Approval Policies to Qraft ⛏︎👮‍♀️

radist2s:
This PR continues the refactoring effort, focusing on removing dependencies on `@monite/sdk-api` from Approval Policies related components.

### Changes
* The main goal of this PR is to remove the usage of `@monite/sdk-api` from Approval Policies components. This includes:
  - Removing imports from `@monite/sdk-api`
  - Replacing API calls with OpenAPI Qraft
  - Updating fixtures and handlers related to Approval Policies
* useApprovalPolicies* (multiple hooks related to Approval Policies)


packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/useApprovalPolicyDetails.tsx
```diff
@@ -26,10 +27,42 @@ export const useApprovalPolicyDetails = ({
...
-  const createMutation = useCreateApprovalPolicy();
-  const updateMutation = useUpdateApprovalPolicy();
+  const { i18n } = useLingui();
+  const { api, queryClient } = useMoniteContext();
+
+  const createMutation = api.approvalPolicies.postApprovalPolicies.useMutation(
+    {},
+    {
+      onSuccess: async () => {
+        await api.approvalPolicies.getApprovalPolicies.invalidateQueries(
```
saladNights, line 37: While you migrated the hook absolutely right, I spotted the error that the detailed approval policy data was not invalidated. The error was in the previous hooks as well.
Could you add the `api.approvalPolicies.getApprovalPoliciesId` invalidation? 
```
await api.approvalPolicies.getApprovalPoliciesId.invalidateQueries(
          {
            parameters: { path: { approval_policy_id: approvalPolicy.id } },
          },
          queryClient
        );
```
The same for `updateMutation`

saladNights (changes requested):

packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/useApprovalPolicyDetails.tsx
radist2s: Thank you, such a nice catch, fixed.

packages/sdk-react/src/components/approvalPolicies/ApprovalPolicyDetails/ApprovalPolicyDetails.test.tsx
```diff
@@ -125,32 +124,18 @@ describe('ApprovalPolicyDetails', () => {
...
-    const callArguments = createApprovalPolicy.mock.calls;
-
-    if (!callArguments.length) {
-      throw new Error('createApprovalPolicy was not called');
-    }
-
-    const requestBody = callArguments[0][0];
-
-    expect(requestBody).toMatchObject({
+    expect(requestFnMock.mock.lastCall?.[1].body).toMatchObject({
```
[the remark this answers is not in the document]
radist2s: This is a valid approach, but in this case, we are rather testing the Mutation Cache. For this test, it is enough to see if the request was called with the correct parameters. If `requestFn` is called, it means that it was called by Qraft (hardly anyone else), and the result with those parameters is guaranteed to be in the Mutation Cache.

saladNights (approved):

PR #162, opened 2024-07-11
feat(DEV-11730): migrate Roles to Qraft

saladNights:
This PR implements a critical architectural change: migrating the Roles component from the legacy @monite/sdk-api to OpenAPI Qraft.

Details
Migration to Qraft:

Replaced all @monite/sdk-api dependencies in the Roles flow with Qraft-generated API clients
Removed packages/sdk-react/src/core/queries/useRoles


packages/sdk-react/src/components/userRoles/UserRoleDetails/helpers.ts
```diff
@@ -65,8 +63,27 @@ const transformActionsToComponentFormat = <T extends string, P>(
...
+): PayablePermissionRow | CommonPermissionRow => {
+  let permission: Record<string, boolean | typeof objectType> = {};
+
+  permission.name = objectType;
+
+  actionEnum.forEach((action) => {
+    permission[action] = false;
+  });
+
+  return permission as PayablePermissionRow | CommonPermissionRow;
```
radist2s, line 82: I see a complicated union, but, seems, output generic is needed. However, if typing is too complicated, this is fine.

My only suggestion would be: that's fine to do `as T` for generics resolution.

In my case, if I  need to apply `as` to something that definitely could be determined using runtime checks, but I didn't do this - here is the point where TypeScript reminds me that, probably, the initial implementation was wrong.

radist2s (approved):
Unbelievable! Brief! On time!

packages/sdk-react/src/components/userRoles/UserRoleDetails/helpers.ts
saladNights: Yes, the types of roles implementation are overcomplicated. I realized it again when I was moving it to Qraft.
I created a task in our tech debt epic.

PR #171, opened 2024-07-17
Fix/dev 11864 counterpart contacts

chebum:


chebum:
This PR implements generation of counterpart contacts.

I noticed that counterparts generation is duplicated in counterparts.ts and generate-payables.ts

PR #174, opened 2024-07-18
fix(DEV-11864): [Demo] Implement Counterpart Contacts generation for demo data to enable Payment Reminders

chebum:
**Objective**

Enhance the demo data generation process in the sdk-demo-with-nextjs-and-clerk-auth project to include the creation of Counterpart Contacts for organization-type Counterparts, specifically to enable the demonstration of Payment Reminders functionality.

**Background**

Currently, our demo data generation process creates Counterparts but does not generate associated Contacts. This limitation prevents us from showcasing the Payment Reminders feature, as reminders are sent to the email address of the counterpart's default contact (stored in the counterpart_contact.email field of the invoice).


examples/with-nextjs-and-clerk-auth/src/lib/monite-api/demo-data-generator/counterparts.ts
```diff
@@ -57,10 +67,21 @@ interface ICounterpartsBuilderOptions {
...
+     *  for the counterparts
+     */
+    enabled: boolean;
+    count: number;
+  };
 }
 
-interface IConterpartsServiceResponse {
-  counterparts: Array<components['schemas']['CounterpartResponse']>;
+interface ICounterpartsServiceResponse {
```
radist2s, line 83: If it's not too much trouble, let's also change the interface names (we had [a proposal](https://monite.atlassian.net/browse/DEV-10062)).

```ts
ICounterpartsServiceResponse => CounterpartsServiceResponse

ICounterpartsServiceResponse => CounterpartsServiceResponse
```

I think the size of PR will not increase significantly.

radist2s (approved):

PR #181, opened 2024-07-22
fix(DEV-11866): extend data generator with reminders

chebum:
This PR enhances the demo data generation process to create both "Before due date" (`/payment_reminders`) and "Overdue" (`/overdue_reminders`) Reminders for Entities, assign these Reminders to specific Invoices, and add a specific CLI command for generating Reminders.


radist2s (approved):
🫶

PR #200, opened 2024-07-27
feat(DEV-11981): invoice Payment & Overdue reminders 🔔

radist2s:
Added functionality to specify _Payment_ and _Overdue_ Reminders during the creation and editing of invoices.

PR #209, opened 2024-07-30
chore(DEV-12119): add ESLint rule for consistent Boolean conversion for 'enabled' property

vkuprin:
This PR introduces a custom ESLint rule that enforces the consistent use of the !! operator instead of the Boolean() function when converting values to a boolean for the enabled property within objects. This rule aims to standardize the boolean conversion practice across the codebase, enhancing code readability and consistency

Inspiration coming from this conversation: https://github.com/team-monite/monite-sdk/pull/200#discussion_r1695703430

It also has auto-fix mode
Screenshot 2024-07-31 at 15 23 10
Screenshot 2024-07-31 at 15 23 16


denis-timin:
> This PR introduces a custom ESLint rule that enforces the consistent use of the !! operator instead of the Boolean() function when converting values to a boolean for the enabled property within objects. This rule aims to standardize the boolean conversion practice across the codebase, enhancing code readability and consistency
> 
> Inspiration coming from this conversation: [#200 (comment)](https://github.com/team-monite/monite-sdk/pull/200#discussion_r1695703430)

@vkuprin : Vitaly, I suggest we use Boolean(...) instead of !! for converting values to booleans. Boolean(...) is more readable and self-explanatory, making our code more approachable, especially for those new to our codebase or JavaScript in general (remember, we have clients reading our code). The performance difference is minimal and not a concern. While !! is a concise shorthand, Boolean(...) prioritizes clarity and maintainability, which are crucial in our case of a public codebase.

So, to sum up, I am 100% for maintaining the uniformity of the repository, we should all strive to improve it, but let's also follow the readability and maintainability.

denis-timin:
@vkuprin : please also add @chebum to reviewers.

vkuprin:
> > This PR introduces a custom ESLint rule that enforces the consistent use of the !! operator instead of the Boolean() function when converting values to a boolean for the enabled property within objects. This rule aims to standardize the boolean conversion practice across the codebase, enhancing code readability and consistency
> > Inspiration coming from this conversation: [#200 (comment)](https://github.com/team-monite/monite-sdk/pull/200#discussion_r1695703430)
> 
> @vkuprin : Vitaly, I suggest we use Boolean(...) instead of !! for converting values to booleans. Boolean(...) is more readable and self-explanatory, making our code more approachable, especially for those new to our codebase or JavaScript in general (remember, we have clients reading our code). The performance difference is minimal and not a concern. While !! is a concise shorthand, Boolean(...) prioritizes clarity and maintainability, which are crucial in our case of a public codebase.
> 
> So, to sum up, I am 100% for maintaining the uniformity of the repository, we should all strive to improve it, but let's also follow the readability and maintainability.

We discussed directly to keep consistency in chat https://github.com/team-monite/monite-sdk/pull/200#discussion_r1695681477 We can use different format Boolean everywhere

radist2s (commented):
I have some concerns about the introduction of a custom ESLint rule for Boolean conversion. Firstly, instead of creating a custom rule, we could simply use the existing [`no-implicit-coercion`](https://eslint.org/docs/latest/rules/no-implicit-coercion) rule, which already handles this scenario effectively.

However, considering that a significant portion of our codebase is JSX, enforcing this rule might lead to issues.

For example, using `Boolean(...)` can cause TypeScript to emit type errors, while `!!` works perfectly fine in JSX:

```tsx
let foo: { bar: 'baz' } | undefined;

return (
  <>
    {!!foo?.bar && foo.bar} // ✅ This works fine
  </>
);
```

On the other hand, using `Boolean(...)` might result in type errors:

```tsx
let foo: { bar: 'baz' } | undefined;

return (
  <>
    {Boolean(foo?.bar) && foo.bar} // 🔴 TypeScript error: foo is possibly undefined
  </>
);
```

This could result in the need for additional `eslint-disable` comments or workarounds, which would only clutter our codebase further.

Moreover, I don't believe that standardizing the use of `Boolean(...)` or `!!` is critical enough. _It might be simpler and more efficient to agree on this practice informally within the team._

chebum:
I believe, the issue isn't important enough to spend anymore time on it at the moment.

PR #213, opened 2024-07-31
chore(proposal): ESLint plugin for unused imports and variables

vkuprin:
This pull request includes the addition of eslint-plugin-unused-imports and the configuration of the ESLint rule no-unused-vars for better handling of unused code.

The changes include:

- Added eslint-plugin-unused-imports: This plugin helps identify and remove unused imports from the codebase.

Updated ESLint Rules:

- no-unused-vars: Configured to warn about unused variables
- unused-imports/no-unused-imports: Flags unused imports as errors, helping to maintain a clean codebase


packages/sdk-api/src/api/services/CommonService.ts
nekedos, line 1: It is an auto-generated file. Is it possible to add it to ignorance?

packages/sdk-api/src/api/services/CurrenciesInternalService.ts
nekedos, line 1: The same. This file is auto-generated.

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartAddressFormUpdate/useCounterpartAddressFormUpdate.tsx
```diff
@@ -27,6 +27,8 @@ export function useCounterpartAddressFormUpdate({
   counterpartId,
   addressId,
   onUpdate,
+  // ToDo: check why we are not using onCancel here, but passing it as prop from parent component
```
nekedos, line 30: You can remove it because it is just object destructuring

packages/sdk-react/src/components/onboarding/OnboardingPersonDocuments/OnboardingPersonDocuments.tsx
```diff
@@ -57,7 +57,7 @@ export const OnboardingPersonDocuments = () => {
 
     if (!fields) return undefined;
 
-    const { id, ...documents } = fields;
```
nekedos, line 60: I don't remember why I added this code.

But in this case, it is better to return `fields ` directly.

vkuprin: Not sure, I understand suggestion. `id` was not used, that's why I removed it

Screenshot 2024-07-31 at 21 43 57

packages/sdk-api/src/api/services/CommonService.ts
vkuprin: That’s an excellent point. I’ve implemented an additional override in our ESLint configuration that appears to work well for our specific case:

```json
{
  "files": ["sdk-api/src/api/**/*"],
  "rules": {
    "unused-imports/no-unused-imports": "off"
  }
}
```
This override specifically ignores only unused imports. I believe it would be unwise to apply a blanket ESLint ignore to all generated files. What are your thoughts on this approach?

vkuprin: Done, added ignorance 

```js
 {
    files: ['packages/sdk-api/src/api/**/*'],
    rules: {
      'unused-imports/no-unused-imports': 'off',
    },
  }
```

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartAddressFormUpdate/useCounterpartAddressFormUpdate.tsx
vkuprin: If we remove this `onCancel`, it would also be logical to eliminate the interface for `onCancel` as well. However, doing so could lead to further complications:

Screenshot 2024-07-31 at 22 07 52

Should we consider removing the onClick functionality entirely? It seems we may have identified a potential flaw in the application

nekedos (approved):

PR #215, opened 2024-07-31
feat(DEV-12074): Migrate to Material Data Grid library v7

saladNights:
All columns in all tables should allow to change their width by dragging one of the edges of a header cell.

MUI Data Grids in v7 allows to change the width of a column out of the box.
So, we decided to migrate to the v7.


packages/sdk-react/src/components/approvalRequests/ApprovalRequestsTable/useApprovalRequestActionsCell.tsx
```diff
@@ -91,7 +91,7 @@ export const useApprovalRequestActionsCell = (
     field: 'actions',
     renderHeader: () => null,
     sortable: false,
-    flex: 0.5,
+    flex: 0.6,
```
vkuprin, line 94: Why did we increase this value by 0.1? Did the migration change the styles?

saladNights: Cells in v7 are not `flex` anymore.
I had to adjust it everywhere when the inner container relied on the outer container's `flex` style.

This particular example wasn't fixed properly. Thanks for the comment, I rewrote it.

chebum (approved):

PR #216, opened 2024-08-01
chore(proposal): upgrade Yarn 3 -> 4

vkuprin:


e2e/bin/lib/update-projects-with-workspace-packages.mjs
```diff
@@ -15,7 +15,7 @@ function getWorkspacesPublishingPackages() {
   const prefixSalt = '[package]:';
 
   const result = execSync(
-    `(cd "${findMonorepoRoot()}" && yarn workspaces foreach --no-private --from '@team-monite/*' --from '@monite/*' \
+    `(cd "${findMonorepoRoot()}" && yarn workspaces foreach --no-private --recursive --from '@team-monite/*' --from '@monite/*' \
```
radist2s, line 18: I would also add `--topological` option to the command line.

package.json
```diff
@@ -52,5 +52,5 @@
     "@types/react": "18.3.1",
     "@types/react-dom": "18.3.0"
   },
-  "packageManager": "yarn@3.5.0"
+  "packageManager": "yarn@4.3.1"
```
radist2s, line 55: It seems that Yarn v3.4.0 is already released.

vkuprin: 3.4.0 is previous version

e2e/bin/lib/update-projects-with-workspace-packages.mjs
vkuprin: Could you please explain why this change is necessary? I believe the current approach addresses the issues. If your suggestion offers significant benefits, I’d appreciate more context and information. Thank you for sharing your knowledge

package.json
radist2s: Misspelling, `4.4.0` of course is out.

e2e/bin/lib/update-projects-with-workspace-packages.mjs
radist2s: I thought it was always better to give a direction [to the source](https://yarnpkg.com/cli/workspaces/foreach#:~:text=If%20%2Dt%2C-,%2D%2Dtopological,-is%20set%2C%20Yarn) than to explain the documentation.

vkuprin: While I understand how to find information in the documentation, I still don’t grasp why we actually need the --topological option in this case. Could you please explain its specific importance and benefits in our workflow?

chebum: @radist2s because we only want to run this command only if all previous commands successfully finished?

radist2s: Nope. Just for the clarity of any future potential issues. With the `--topological` first you publish:
1. `@monite/sdk-api`
2. `@monite/sdk-react`

For now, it's not a conflict, I don't mind if you keep it as is. At the same time, I see no issue with just adding the option.

vkuprin: Thanks, I will add this in that case

radist2s (approved):
🫶

PR #222, opened 2024-08-02
fix(DEV-12155): fix Amount field being clipped in Receivables / Payables data grids

chebum:
This PR fixes Amount field being clipped in Receivables / Payables data grids. 
The implementation uses setTimeout for calls synchronisation due to poor API of the MUI DataGrid.

Code was adapted from https://mui.com/x/react-data-grid/column-dimensions/#autosizing-asynchronously to work with hooks and functional components.

Docs say:
"The Data Grid can only autosize based on the currently rendered cells. DOM access is required to accurately calculate dimensions"
Which means that autosize function should be called after DataGrid renders fully.


packages/sdk-react/src/components/receivables/CreditNotesTable/CreditNotesTable.tsx
```diff
@@ -76,6 +77,29 @@ const CreditNotesTableBase = ({ onRowClick }: CreditNotesTableProps) => {
...
 
+  // Adapted from https://mui.com/x/react-data-grid/column-dimensions/#autosizing-asynchronously
+  // setTimeout and flushSync are necessary for call order control
+  // Docs say:
+  // The Data Grid can only autosize based on the currently rendered cells.
+  // DOM access is required to accurately calculate dimensions
+  const gridApiRef = useGridApiRef();
+  useEffect(() => {
+    setTimeout(() => {
+      ReactDOM.flushSync(() => {
```
vkuprin, line 88: Consider putting this logic in a separate hook, as we have some duplication here:

```ts
const useAutoSizeColumns = (gridApiRef, dependencies = []) => {
  useEffect(() => {
    const handleAutoSizeColumns = () => {
      setTimeout(() => {
        ReactDOM.flushSync(() => {
          setTimeout(() => {
            if (gridApiRef.current?.autosizeColumns) {
              gridApiRef.current.autosizeColumns({
                columns: ['amount'],
                includeHeaders: true,
                includeOutliers: true,
              });
            }
          }, 1);
        });
      }, 1);
    };

    handleAutoSizeColumns();
  }, [gridApiRef, ...dependencies]);
};
```

chebum: Good point! Thanks! I've just pushed a code refactoring.

vkuprin (approved):

PR #227, opened 2024-08-05
fix(DEV-12024): fix persons step invalid_argument and styles

hypsteria:
fix persons step error, when we don't have country info for user 

image

also fixed styles in table

### DIFF
--- a/packages/sdk-react/src/components/onboarding/OnboardingPersonsReview/OnboardingAddressView/OnboardingAddressView.tsx
+++ b/packages/sdk-react/src/components/onboarding/OnboardingPersonsReview/OnboardingAddressView/OnboardingAddressView.tsx
@@ -17,7 +17,7 @@ export function OnboardingAddressView({
 
   const countryField = {
     required: !!country?.required,
-    value: t(i18n)`${getRegionName(country?.value as AllowedCountries)}`,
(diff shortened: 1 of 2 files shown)


packages/sdk-react/src/components/onboarding/OnboardingPersonsReview/OnboardingAddressView/OnboardingAddressView.tsx
```diff
@@ -17,7 +17,9 @@ export function OnboardingAddressView({
 
   const countryField = {
     required: !!country?.required,
-    value: t(i18n)`${getRegionName(country?.value as AllowedCountries)}`,
+    value:
+      country?.value &&
+      t(i18n)`${getRegionName(country.value as AllowedCountries)}`,
```
nekedos, line 22: It is better to remove `as`:

```ts
value: country?.value ? t(i18n)`${getRegionName(country?.value)}` : '',
```

hypsteria: updated

radist2s (approved):

radist2s (approved):
Thank you very much for the fix! ❤️‍🩹

vkuprin (approved):

PR #234, opened 2024-08-08
fix(DEV-11940): fix toasts for common erorrs

hypsteria:
Show toast for common errors in onboarding form (not only errors in fields)

Screenshot 2024-08-08 at 10 20 53 AM

### DIFF
--- a/packages/sdk-react/src/components/onboarding/hooks/useOnboardingPerson.ts
+++ b/packages/sdk-react/src/components/onboarding/hooks/useOnboardingPerson.ts
@@ -10,6 +10,7 @@ import {
 import { useOnboardingRequirementsContext } from '../context';
 import { isRepresentative } from '../helpers';
 import { enrichFieldsByValues } from '../transformers';
+import { showErrorToast } from '../utils';
 import { useOnboardingActions } from './useOnboardingActions';
 import { useOnboardingEntity } from './useOnboardingEntity';
 import type {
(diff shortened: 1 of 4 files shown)


packages/sdk-react/src/components/onboarding/hooks/useOnboardingPerson.ts
```diff
@@ -77,15 +78,19 @@ export function useOnboardingPerson(): OnboardingPersonReturnType {
     isPending: isCreateLoading,
     error: createPersonError,
   } = api.persons.postPersons.useMutation(undefined, {
-    onError: () => {},
+    onError: (error) => {
+      showErrorToast(error);
```
radist2s, line 82: We have a special function for that: `getAPIErrorMessage(...)`.

hypsteria: @radist2s getAPIErrorMessage shows every message from backend. We use new function to show toasts only in cases when we have error.message. In cases when we have error.details we show errors in fields. Initially I set this function empty `onError: () => {}` to not to show any toasts, but we need this toasts for common errors

radist2s: Sorry, thanks, didn't realise that from the start. Can I ask you to add a comment to the function to make it clear what the exception is for?

radist2s (approved):

packages/sdk-react/src/components/onboarding/hooks/useOnboardingPerson.ts
hypsteria: @radist2s yes, thanks. Added description to showErrorToast.ts

PR #236, opened 2024-08-09
feat(DEV-12073): status chip colors

chebum:
This PR adds chip colors according to Figma.


examples/with-nextjs-and-clerk-auth/src/themes/monite.ts
```diff
@@ -290,6 +320,104 @@ const components: Components<Omit<Theme, 'components'>> = {
...
+        props: { status: 'draft' },
+        style: statusColors.black,
+      },
+      {
+        props: { status: 'issued' },
+        style: statusColors.blue,
+      },
+      {
+        props: { status: 'accepted' },
+        style: statusColors.green,
```
radist2s, line 340: In my opinion, it doesn't make sense to put colors in a separate variable that has abstract keys "green", "blue" and so on. It is much more clear when colors are simply duplicated because the IDE can see at a glance which color is set for which token.

chebum: It's how I started implementing them. Unfortunately, the code quickly became difficult to comprehend because we have lots of statuses. Extracting colors to separate variables made the code easier to understand what it does.

radist2s (approved):

PR #241, opened 2024-08-10
fix(DEV-12191): allow grid to be wider that page width

chebum:
Goal of this PR is to
- User should be able to resize DataGrid to be wider than visible screen width.
- User should be able to change columns width
- Column width should be preserved when sorting or changing pages
- Column widths should be selected to display all data without clipping


packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
```diff
@@ -10,22 +10,41 @@ import { useGridApiRef } from '@mui/x-data-grid';
...
-              columns: ['amount'],
+    const grid = gridApiRef.current;
+    if (!grid || !grid.autosizeColumns) return;
+    if (columnWidths && columnWidths.length > 0) {
+      for (const column of columnWidths) {
+        grid.setColumnWidth(column.field, column.width ?? 100);
+      }
+    } else {
+      setTimeout(() => {
+        ReactDOM.flushSync(() => {
```
radist2s, line 26: I recommend changing `*` imports to named imports:
```ts
import { flushSync } from 'react-dom';
```

chebum: don

chebum:
1. Extract columns to useMemo hook to preserve column definition instances between re-renders - this solves collapsing columns between re-renders.
2. Perform column auto-size when columns set changes.
3. useCurrencies.formatCurrencyToDisplay wrapped into useCallback to preserve function instance between re-renders.

packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx
```diff
@@ -161,7 +165,133 @@ const PayablesTableBase = ({
     }
   }, [isError, error, i18n]);
 
-  const gridApiRef = useAutosizeGridColumns(payables);
+  const columns = useMemo<GridColDef[]>(() => {
+    return [
+      {
+        field: 'document_id',
```
saladNights, line 171: you removed display: flex for columns 'document_id' and 'created_at'.
It breaks vertical spacing for the payable in OCR status (see screenshot).
Bring back display: flex if possible. Otherwise, please find a way to center icons vertically (probably wrap them with flex)

Screenshot 2024-08-15 at 20 32 35

saladNights (changes requested):

packages/sdk-react/src/components/approvalRequests/ApprovalRequestsTable/ApprovalRequestsTable.tsx
```diff
@@ -165,6 +165,76 @@ const ApprovalRequestsTableBase = ({
...
+          <ApprovalRequestStatusChip status={params.value} />
+        ),
+      },
+      {
+        field: 'amount',
+        headerName: t(i18n)`Amount`,
+        sortable: false,
+        flex: 0.5,
+        valueGetter: (_, row) => {
+          const payable = row;
```
vkuprin, line 220: **Small Suggestions:**

No need to create the payable variable here. You can directly access and mutate row.amount_to_pay and row.currency without the extra step, simplifying the code and avoiding unnecessary memory allocation.

```ts
valueGetter: (_, row) => {
  return row.amount_to_pay && row.currency
    ? formatCurrencyToDisplay(row.amount_to_pay, row.currency)
    : '';
},
```

Also, consider applying this approach in other valueGetter logic. I’ve noticed multiple instances where variables are allocated without adding extra value

If you’re not performing additional steps with the variable, try mutating directly

saladNights (approved):

PR #242, opened 2024-08-11
fix(DEV-12158): use flex for datagrid height instead of absolute unit px

saladNights:
Styles for table components were updated and autoHeight property was removed.
Now all tables reproduce the default MUI DataGrid behavior, they require a parent container with intrinsic dimensions.
Advantages:

- we don't force extra property (autoHeight). This property is not recommended in some cases
- the table component could be embedded in a parent block with absolute or dynamic height. It will grow till the bottom of the parent container and leave the content area scrollable. Examples are in the Storybook and Demo project.

### DIFF
--- a/packages/sdk-react/src/components/approvalPolicies/ApprovalPoliciesTable/ApprovalPoliciesTable.tsx
+++ b/packages/sdk-react/src/components/approvalPolicies/ApprovalPoliciesTable/ApprovalPoliciesTable.tsx
@@ -133,101 +133,104 @@ const ApprovalPoliciesTableBase = ({
   };
 
   return (
-    <>
-      <Box
-        sx={{ padding: 2, width: '100%', height: '100%' }}
-        className={ScopedCssBaselineContainerClassName}
-      >
-        <Box sx={{ marginBottom: 2 }}>
-          <Filters onChangeFilter={onChangeFilter} />
-        </Box>
-        <DataGrid
-          autoHeight
-          disableColumnFilter={true}
-          rowSelection={false}
-          sx={{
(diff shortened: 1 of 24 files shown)


chebum (changes requested):
Counterparts table still grows uncontrollably

chebum (approved):
Thank you for fixing the page layout - SDK Demo looks way better now.

packages/sdk-react/src/components/approvalPolicies/ApprovalPoliciesTable/ApprovalPoliciesTable.tsx
```diff
@@ -133,101 +133,104 @@ const ApprovalPoliciesTableBase = ({
...
+      <Box sx={{ mb: 2 }}>
+        <Filters onChangeFilter={onChangeFilter} />
       </Box>
-    </>
+      <DataGrid
+        disableColumnFilter={true}
+        rowSelection={false}
+        sx={{
+          '&.MuiDataGrid-root--densityCompact .MuiDataGrid-cell': {
+            py: '8px',
```
radist2s, line 154: In my opinion, it would be better to use something like `py: 2` to keep internal spacing consistent.

chebum: @radist2s I agree. Selector '&.MuiDataGrid-root--densityCompact .MuiDataGrid-cell' itself isn't great though. I wouldn't worry about hardcoded 8px until this selector moved to a theme file.
This refactoring is out of scope of this PR though.

PR #244, opened 2024-08-14
feat(proposal): improve SDK Demo Authorization form 🧽

radist2s:
- Moved `<EntityIdLoader/>` from `@monite/sdk-drop-in` to `@team-monite/sdk-demo`.
  - (`@monite/sdk-drop-in` is bundled, therefore changesets are unnecessary)
  - `entity_id` is not needed anymore in the form 🥂
- Improved SDK Demo authorization flow by enhancing error handling.
- Prevented local storage from being cleared when authorization fails.

<details>
  <summary>Error Message screenshot with Logout button</summary>
  image
</details>


packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
```diff
@@ -78,45 +87,71 @@ const SDKDemoComponent = ({
...
+  const fetchToken = useCallback(async () => {
+    try {
+      if (!authData) throw new Error('Auth data is not provided');
+      if (authData instanceof Error) throw authData;
+      const token = await fetchTokenBase(apiUrl, authData);
+      setFetchTokenError(null);
+      return token;
+    } catch (error) {
+      setFetchTokenError(
+        error instanceof Error ? error : new Error(JSON.stringify(error))
```
vkuprin, line 100: Instead of using `JSON.stringify(error)`, which may not always produce a meaningful message, consider using `String(error)` to convert the error to a string

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
```diff
@@ -136,3 +171,44 @@ const MoniteReactQueryDevtools = () => {
...
+        margin="auto"
+        height="100vh"
+        justifyContent="center"
+      >
+        {errors.map((error) => (
+          <Alert severity="error" key={error.message}>
+            Error: <code>{error.message}</code>
+          </Alert>
+        ))}
+        {Boolean(errors.length) && (
```
vkuprin, line 198: The `Boolean(errors.length)` check is redundant here, due to the initial check before `if (!errors.length) return null;`. You can safely remove this conditional rendering

packages/sdk-demo/src/components/EntityIdLoader.tsx
```diff
@@ -88,4 +90,29 @@ export const createEntityUsersMyEntityRequestFn = (
...
+    super(props);
+    this.state = { isError: false };
+  }
+
+  static getDerivedStateFromError(error: unknown) {
+    return { isError: true, error: error };
+  }
+
+  render() {
+    if (this.state.isError) return null;
```
vkuprin, line 113: It’s generally better to provide some fallback UI instead of just returning null

packages/sdk-demo/src/core/fetchToken.ts
```diff
@@ -24,14 +24,13 @@ export async function fetchToken(
   });
 
   if (!res.ok) {
-    throw new Error('Could not fetch token');
+    throw new Error(`Could not fetch token: ${await res.text()}`);
```
vkuprin, line 27: It might return empty string from that promise and the sentence could make no sense ending with " : "

```ts
const errorMessage = await res.text();
throw new Error(`Could not fetch token: ${errorMessage || 'Unknown error'}`);
```

vkuprin (changes requested):
Thanks!

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
radist2s: `String(error)` returns `'[object Object]'` while `JSON.stringify` returns a verbose output.

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
radist2s: Thanks!

packages/sdk-demo/src/components/EntityIdLoader.tsx
radist2s: Yep, I was thinking about that, but could not figure out what exactly.

packages/sdk-demo/src/core/fetchToken.ts
radist2s: In our case, it can't be an empty string.

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
vkuprin: Actually, String(error) only returns "[object Object]" if error is a plain object, not an actual Error instance. If it’s an Error object, it should return something like "Error: Some error". When you call String() on error message .toString is invoked

JSON.stringify(error) does give a more detailed output, but it might miss some properties of Error objects, like message or stack, unless those are explicitly enumerable.

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
```diff
@@ -136,3 +184,42 @@ const MoniteReactQueryDevtools = () => {
...
+      <Stack
+        spacing={2}
+        maxWidth="sm"
+        margin="auto"
+        height="100vh"
+        justifyContent="center"
+      >
+        {errors.map((error) => (
+          <Alert severity="error" key={error.message}>
+            Error: <code>{error.message}</code>
```
saladNights, line 208: add localisation pls

saladNights:
Add a changeset

saladNights (approved):

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
radist2s: Didn't want to inflate the size of the dictionary with util messages. Added.

packages/sdk-demo/src/apps/SDKDemo/SDKDemo.tsx
radist2s: 1. It's a demo project
2. error handling is tested in real cases

vkuprin (approved):

PR #245, opened 2024-08-14
fix(DEV-11417): update `CommonOperator` type according to OpenAPI Document

radist2s:
* Updated the `CommonOoperator.action` property. Previously, the action was incorrectly listed as `PayableActionEnum`, but it should be simply `ActionEnum`.
* Additionally, fixed the handling of these permissions for the Invoice Cancellation button.


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/useExistingInvoiceDetails.tsx
```diff
@@ -48,12 +48,6 @@ export function useExistingInvoiceDetails({
     entityUserId: receivable?.entity_user_id,
   });
 
-  const { data: isCancelAllowed } = useIsActionAllowed({
```
vkuprin, line 51: Are we sure we don't need this? I included it to check for permissions. Am I missing something, or is this check unnecessary for us?

vkuprin: I just checked out this branch and tested it in Storybook, but I don’t see the Cancel button for the Overdue and Issued statuses, even though it should be there. Am I missing something?

Screenshot 2024-08-15 at 00 12 00
Screenshot 2024-08-15 at 00 12 13

radist2s: Thanks, I'll check.

radist2s: Found one more issue and fixed with `entity?.receivable_edit_flow !== 'compliant'`

vkuprin (approved):

saladNights (approved):

saladNights:
add a changeset

PR #248, opened 2024-08-16
fix(DEV-12296): move monite theme to sdk themes

chebum:
This PR moves Monite theme into SDK-themes from Demo. This allows us to share the theme between Demo and IFrame deployments.


chebum:
@saladNights  Per suggestions, I removed all .d.ts files from sdk-themes and demo apps.
I added the capability to specify color values by typing path in the palette object: 'divider', 'primary.main', 'neutral.50', etc.
I copied some code from @mui/utils since we cannot import external modules in the sdk-themes package.

examples/with-nextjs-and-clerk-auth/src/components/Layout.tsx
```diff
@@ -20,12 +20,8 @@ export const Layout = ({ children }: { children: ReactNode }) => {
       display="flex"
       className={`Monite-Page-${currentPath}`}
       sx={{
-        position: 'absolute',
-        left: '0',
-        right: '0',
-        top: '0',
-        bottom: '0',
-        overflow: 'hidden',
+        width: '100vw',
```
radist2s, line 23: This is not really the [same](https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser). We could use `lvh` instead, but firefox doesn't support it yet.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
```diff
@@ -209,26 +209,24 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
...
-                  disabled={createReceivable.isPending}
-                  onUpdateOverdueReminder={onEditOverdueReminder}
-                  onUpdatePaymentReminder={onEditPaymentReminder}
-                  onCreateReminder={onCreateReminder}
-                />
-              </Stack>
+            <Typography variant="h1" sx={{ mb: 7 }}>
+              {t(i18n)`Create Invoice`}
+            </Typography>
+            <Stack direction="column" spacing={4} sx={{ pb: 7 }}>
```
radist2s, line 215: I would prefer to use `margin` instead of `padding`, or event `gap` on the parent `<form/>` element.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
```diff
@@ -209,26 +209,24 @@ const CreateReceivablesBase = (props: InvoiceDetailsCreateProps) => {
...
-                />
-                <PaymentSection disabled={createReceivable.isPending} />
-                <ReminderSection
-                  disabled={createReceivable.isPending}
-                  onUpdateOverdueReminder={onEditOverdueReminder}
-                  onUpdatePaymentReminder={onEditPaymentReminder}
-                  onCreateReminder={onCreateReminder}
-                />
-              </Stack>
+            <Typography variant="h1" sx={{ mb: 7 }}>
```
radist2s, line 212: It seems, the margin should be applied to the `<form/>` if we're talking about semantics

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -229,28 +229,23 @@ const EditInvoiceDetailsContent = ({
...
-                />
-                <PaymentSection disabled={isLoading} />
-                <ReminderSection
-                  disabled={isLoading}
-                  onUpdateOverdueReminder={onEditOverdueReminder}
-                  onUpdatePaymentReminder={onEditPaymentReminder}
-                  onCreateReminder={onCreateReminder}
-                />
-              </Stack>
+            <Typography variant="h1" sx={{ mb: 7 }}>
```
radist2s, line 232: Same here, heading should not care about if there is a free space required or not.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -186,7 +186,7 @@ const EmailInvoiceDetailsBase = ({
           </Grid>
         </Toolbar>
       </DialogTitle>
-      <DialogContent className={className + '-Content'}>
+      <DialogContent className={className + '-Content'} sx={{ mt: 4 }}>
```
radist2s, line 189: In case, if we have a custom `className`, why should we add hardcoded margins?

packages/sdk-themes/src/themes/monite.ts
```diff
@@ -182,22 +659,221 @@ export const defaultMoniteComponents: Components<Omit<Theme, 'components'>> = {
...
+    // noinspection SuspiciousTypeOfGuard
+    if (temporaryValue && typeof temporaryValue === 'string')
+      return temporaryValue;
+  } else if (typeof palette[strVal as keyof PaletteOptions] === 'string')
+    return palette[strVal as keyof PaletteOptions] as string;
+  return strVal;
+}
+
+// Replaces color constants like 'divider', 'primary.main', 'neutral.80' with actual color values
+function renderColors<T extends { [key: string]: any }>(
```
radist2s, line 838: Am I correct in understanding that the MUI does not support `primary.main` values in the theme?

examples/with-nextjs-and-clerk-auth/src/components/Layout.tsx
chebum: I removed absolute positioning to preserve consistency across SDK. Since our UI isn't mobile-friendly yet, I believe we can postpone it.

packages/sdk-drop-in/src/apps/MoniteIframeAppDemo.tsx
```diff
@@ -165,7 +165,12 @@ const MoniteIframe = ({
       key={component}
       ref={setIframeElement}
       src={iframeUrl}
-      style={{ border: 'none', width: '100%', height: '100%' }}
+      style={{
+        border: 'none',
+        width: 'calc(100% - 64px)',
```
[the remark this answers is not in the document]
chebum: @radist2s Good catch. Actually, <iframe> should have internal padding. If padding of our tables is provided by iframe container, then form popups inside iframe will have double padding. I reverted the code to previous version.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
chebum: @radist2s Sure. I changed `padding` to `margin`. Gap won't work in this case since we need space after editors.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
chebum: @radist2s I moved the margin to form.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
chebum: @radist2s Changed the code as requested.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
chebum: @radist2s Moving the spacing to the theme will break layout for Material UI theme. We haven't dropped it yet.

packages/sdk-themes/src/themes/monite.ts
chebum: @radist2s Yep. It leaves string values as-is in MUI v7.

packages/sdk-themes/src/themes/monite.ts
```diff
@@ -182,22 +659,221 @@ export const defaultMoniteComponents: Components<Omit<Theme, 'components'>> = {
...
+      prototype === Object.prototype ||
+      Object.getPrototypeOf(prototype) === null) &&
+    !(Symbol.toStringTag in item) &&
+    !(Symbol.iterator in item)
+  );
+}
+
+// Replaces color constant like 'divider', 'primary.main', 'neutral.80' with actual color value
+function renderColor(strVal: string, palette: PaletteOptions): string {
+  if (strVal.includes('.')) {
```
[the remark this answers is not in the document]
chebum: @radist2s I agree with that. I hardcoded a list of properties that will be processed: 'color', 'backgroundColor', 'bgcolor', 'borderColor'

radist2s (approved):

PR #252, opened 2024-08-19
feat(DEV-12283): payable upload file support pdf | png | jpeg | tiff

vkuprin:


packages/sdk-react/src/components/payables/Payables.tsx
```diff
@@ -159,7 +159,8 @@ const PayablesBase = ({
       )}
 
       <FileInput
-        accept="application/pdf"
+        // eslint-disable-next-line lingui/no-unlocalized-strings
```
radist2s, line 162: I would suggest extending [ESLint rule](https://github.com/team-monite/monite-sdk/blob/b6fd91a927cdf0cd2e6b8e9158832c8309090315/packages/eslint-plugin/src/config/lingui.js#L27)

radist2s (approved):

PR #253, opened 2024-08-19
feat(DEV-12181): configure custom fields that need to be mandatory after OCR for user to check

vkuprin:
- Create a logic that will check all the fields that weren’t filled in the OCR scan
- TBD - mark required fields in red
- Ability for the dev to pass in configs which will mark the required fields that are necessary for OCR. For example, pass in to <PayableDetails> a prop that will be called ocr_required which will have the payload:


packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -562,15 +550,10 @@ const PayableDetailsFormBase = forwardRef<
                                     InputLabelProps={{
                                       style: {
                                         color: isFieldRequired('tags')
-                                          ? 'red'
```
radist2s: Gosh! 😵

vkuprin: Well, The PR was in draft

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -137,6 +142,16 @@ export const PayableDetailsForm = forwardRef<
   </MoniteScopedProviders>
 ));
 
+export const defaultRequiredFields = {
+  invoiceNumber: true,
+  dueDate: true,
+  tags: true,
+  currency: true,
+  invoiceDate: false,
```
chebum, line 150: Fields with 'false' can be dropped in code since they aren't required and this won't affect the code behavior.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -234,6 +252,16 @@ const PayableDetailsFormBase = forwardRef<
 
     const className = 'Monite-PayableDetailsForm';
 
+    const isFieldRequired = (fieldName: keyof typeof defaultRequiredFields) => {
```
chebum, line 255: A very similar function with the same name can be found in PayableDetailsInfo.tsx. Can we keep only one function?

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -154,7 +175,16 @@ const PayableDetailsInfoBase = ({
...
               <TableBody>
                 <TableRow>
-                  <StyledLabelTableCell>
+                  <StyledLabelTableCell
+                    style={{
+                      color: isFieldRequired(
+                        'invoiceNumber',
+                        payable.document_id
+                      )
+                        ? theme.palette.error.main
```
chebum, line 184: A couple of comments here:
1. We're coupling our code with a child component's implementation here.
2. This code repeats itself in several places.

We can deduplicate the code by creating and new class like

    interface StyledLabelTableCellProps {
      children?: React.ReactNode;
      isError?: boolean;
    }
    const StyledLabelTableCell = (props: StyledLabelTableCellProps) => {
      const theme = useTheme();
      return <TableCell sx={{
        color: props.isError ? theme.palette.error.main : theme.palette.text.secondary,
        minWidth: 120,
        width: '35%', 
      }}>{props.children}</TableCell>;
    };

chebum (changes requested):

packages/sdk-react/src/components/payables/PayableDetails/PayableDetails.tsx
```diff
@@ -196,12 +198,14 @@ const PayableDetailsBase = ({
...
                   lineItems={lineItems}
                   payableDetailsFormId={payableDetailsFormId}
+                  ocrRequiredFields={ocrRequiredFields}
                 />
               ) : (
                 payable && (
                   <PayableDetailsInfo
                     payable={payable}
                     optionalFields={optionalFields}
+                    ocrRequiredFields={ocrRequiredFields}
```
radist2s, line 208: @saladNights, do you think it might make more sense to use `Theme` and customize `PayableDetailsInfo` here?
I wouldn't really want to drag this customization into the public `PayableDetails` interface, especially since it would be impossible to customize this for Drop-In or Iframe.

chebum: that's a good idea. It may not be perfect since it's a mix of concepts, but we already have mechanism in place.

saladNights: I vote for customization via theme.

Moreover, I would move the `optionalFields` in the future major version well.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
radist2s: Nope, the comment is for the **deleted** line. Hardcoded `red`... 🤮

vkuprin: Ahh, true 😂

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -234,6 +238,12 @@ const PayableDetailsFormBase = forwardRef<
 
     const className = 'Monite-PayableDetailsForm';
 
+    const ocrFields = useOcrFields();
+
+    useEffect(() => {
+      trigger();
```
radist2s, line 244: I wouldn't want to pull validation for all fields on a form as soon as we go to it without additional conditions.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -248,9 +293,13 @@ const PayableDetailsInfoBase = ({
                 </TableRow>
                 {showTags && payable.tags && payable.tags.length > 0 && (
                   <TableRow>
-                    <StyledLabelTableCell>
+                    <RequiredFieldLabel
+                      fieldName="tags"
+                      ocrRequiredFields={ocrFields}
```
radist2s, line 298: Probably, would be better to replace `ocrRequiredFields` with `required` property.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -64,6 +75,26 @@ const StyledLabelTableCell = styled(TableCell)(({ theme }) => ({
   width: '35%',
 }));
 
+const RequiredFieldLabel = ({
```
radist2s, line 78: I would prefer to keep all child components in the end of the file. Its much easier when you read the code to start with the main parts.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -225,3 +226,30 @@ function formatTaxToMinorUnits(tax: number): number {
...
+
+  const isDefaultRequired = defaultRequiredFields[fieldName] ?? false;
+  const isOcrRequired = ocrRequiredFields?.[fieldName] ?? false;
+
+  return isDefaultRequired || isOcrRequired;
+};
+
+export const useOcrFields = () => {
+  const theme = useTheme();
+  return theme.components?.MonitePayableDetailsInfo?.defaultProps || {};
```
chebum, line 254: @vkuprin Shouldn't it be theme.components?.MonitePayableDetailsInfo?.defaultProps?.ocrFields

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -225,3 +226,30 @@ function formatTaxToMinorUnits(tax: number): number {
 function formatTaxFromMinorUnits(tax: number): number {
   return tax / 100;
 }
+
+export const isFieldRequired = (
+  fieldName: string,
+  ocrRequiredFields: Record<string, boolean> | undefined,
+  value?: string | null
+) => {
```
radist2s, line 234: I would add some generics on it to have strictly typed `fieldName` and `ocrRequiredFields`

packages/sdk-react/mui-styles.d.ts
```diff
@@ -41,6 +42,7 @@ declare module '@mui/material/styles' {
     MonitePayableStatusChip: Partial<PayableStatusChipProps>;
     MoniteApprovalRequestStatusChip: Partial<MoniteApprovalRequestStatusChipProps>;
     MoniteTablePagination: Partial<MoniteTablePaginationProps>;
+    MonitePayableDetailsInfo: Partial<PayableDetailsInfoProps>;
```
chebum, line 45: I can't find a definition of the PayableDetailsInfoProps interface. Probably, forgot to add to the repo?

radist2s (changes requested):

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
chebum: There is a hook called useThemeProps that's better suited for this use case. You can find an example in InvoiceStatusChip

chebum (changes requested):

radist2s:
We should add a TSDoc for the `<PayableDetailsForm/>` [customization](https://github.com/team-monite/monite-sdk/blob/3868d7aac1db82fd6347f36983b1eb710a5a7107/packages/sdk-react/src/components/receivables/ReceivablesTable/ReceivablesTable.tsx#L59-L123), and according [story](https://github.com/team-monite/monite-sdk/blob/3868d7aac1db82fd6347f36983b1eb710a5a7107/packages/sdk-react/src/components/receivables/ReceivablesTable/ReceivablesTable.stories.tsx#L31-L89).

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -225,3 +226,30 @@ function formatTaxToMinorUnits(tax: number): number {
...
+  const defaultRequiredFields: Record<string, boolean> = {
+    invoiceNumber: true,
```
radist2s, line 236: I would really recommend naming the fields in the same way as they are specified in OpenAPI. Otherwise, it would be painful to document.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
vkuprin: My intention here is to trigger this action so that it highlights any fields that are missing for the user. This was specifically requested by @arthemium  to ensure proper highlighting when fields are incomplete

vkuprin: Discussed directly with @radist2s , we want to trigger only OCR fields not all fields

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
chebum: @radist2s Alex, probably, we should create a pseudo component MoniteThemeSettings that will allow us to pass settings like ocrFields and date format?

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -520,3 +558,23 @@ const PayableCounterpartName = ({
     </Stack>
   );
 };
+
+const RequiredFieldLabel = ({
```
radist2s, line 562: I would extend existing `StyledLabelTableCell`. You can just convert it to a normal component or use [`InvoiceStatusChip`](https://github.com/team-monite/monite-sdk/blob/1fd6719da4d2c73b7680afb78b1697bda098e4ca/packages/sdk-react/src/components/receivables/InvoiceStatusChip/InvoiceStatusChip.tsx#L86) as an example.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -303,7 +347,10 @@ const PayableDetailsFormBase = forwardRef<
                             fullWidth
                             error={Boolean(error)}
                             helperText={error?.message}
-                            required
+                            required={isFieldRequired(
```
radist2s, line 350: You have to mark fields as required only if `payable.ocr_status` is empty (it's `null`, but OpenAPI doesn't support nulls). For normal payables this check is not needed.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -234,6 +268,16 @@ const PayableDetailsFormBase = forwardRef<
...
     const className = 'Monite-PayableDetailsForm';
 
+    const { ocrRequiredFields } = useOcrFields(inProps);
+
+    useEffect(() => {
+      const isRequiredFieldMissed = ocrRequiredFields
+        ? Object.values(ocrRequiredFields).some((required) => required)
+        : false;
+
+      if (isRequiredFieldMissed) trigger();
```
radist2s, line 278: I would simplify the logic as `ocrFieldsToTrigger.forEach(field => trigger('field', field)`

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+const payableId = '9a3b97a5-a1ba-4d8c-bade-ad3c47ae61e0';
+
+type Story = StoryObj<typeof PayableDetailsForm>;
+
+const actions = {
+  onSavePayable: action('onSavePayable'),
+  onCreatePayable: action('onCreatePayable'),
+};
+
+const WithDialogComponent: FC<PayableDetailsFormProps> = (args) => {
```
radist2s, line 26: I see no point in making the component as an external, since it is a part of the story.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+      open={open}
+      fullScreen
+      onClose={() => {
+        action('onClose')();
+        setOpen(false);
+      }}
+      onSubmit={() => {}}
+      onClosed={action('onClosed')}
+    >
+      <PayableDetailsForm {...args} />
```
radist2s, line 40: The story implementation does not show how could we configure these fields.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  PayableDetailsForm,
+  PayableDetailsFormProps,
+} from './PayableDetailsForm';
+
+const meta: Meta<typeof PayableDetailsForm> = {
+  title: 'Payables/Payables — Details Form',
+  component: PayableDetailsForm,
+};
+
+const payableId = '9a3b97a5-a1ba-4d8c-bade-ad3c47ae61e0';
```
radist2s, line 17: I would directly use fixture identifier instead of hardcoding. It's hard to maintain.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -225,3 +229,30 @@ function formatTaxToMinorUnits(tax: number): number {
...
+  TFieldValues extends FieldValues
+>(
+  fieldName: TFieldNames,
+  ocrRequiredFields: TOcrRequiredFields | undefined,
+  value?: FieldValue<TFieldValues>
+) => {
+  if (value) return false;
+
+  return (
+    ['invoiceNumber', 'dueDate', 'tags', 'currency'].includes(fieldName) ||
```
radist2s, line 245: By default, we have no predefined fields.

radist2s (dismissed):

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
vkuprin: We have fields that are marked as required before, they were predefined

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
vkuprin: I took this ID from the test cases. Let’s handle it in a separate PR. If you suggest changing the hardcoded values, it’s a good idea to apply this change across the entire project

Screenshot 2024-08-23 at 14 23 18

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
vkuprin: Could you explain what’s missing? I don’t quite understand what changes I need to make

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -520,3 +558,23 @@ const PayableCounterpartName = ({
...
+  fieldName,
+  required,
+  fieldValue,
+  children,
+}: RequiredFieldLabelProps) => {
+  const theme = useTheme();
+  const isRequired = isFieldRequired(fieldName, required, fieldValue);
```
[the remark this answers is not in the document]
vkuprin: I don’t understand your point. The current logic solves the problem. Could you clarify at what level you’re talking about and which required property you’re referring to? I appreciate your suggestions

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
vkuprin: We have fields that are marked as required before, they were predefined. Can you please navigate, what we should do in this case?

radist2s: The idea is to keep backward compatibility as full as possible. This new functionality could be easily reached with the options in the theme. So, by default, we should not display these fields as required for existing customers until they configured them.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
vkuprin: Okey, I will do this, but modular vs in scoped components is subjective topic

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
vkuprin: I’ve added the change you suggested. I noticed that we previously had default required fields. Could you help me understand why we need to change them?

Also this is coming from your sugestion:
Screenshot 2024-08-23 at 21 53 22

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsInfo/PayableDetailsInfo.tsx
```diff
@@ -38,17 +42,32 @@ import {
...
 import { OptionalFields } from '../../types';
 import { isPayableInOCRProcessing } from '../../utils/isPayableInOcr';
 import { usePayableDetailsInfo } from './usePayableDetailsInfo';
 
+/**
+ * Payable Details props for MUI theming
+ */
+export interface MonitePayableDetailsInfoProps {
+  /** The Required OCR Fields */
+  ocrRequiredFields?: Record<string, boolean> | undefined;
```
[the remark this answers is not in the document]
vkuprin: You want me to change that type to this?
Screenshot 2024-08-23 at 22 13 41

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
vkuprin: Are you referring to the OCR fields? So, do I need to set up the entire story with MUI extended support just for these specific fields? I’m concerned because we’re on a tight deadline, and spending time on something that may not be necessary could slow us down. Are you sure this is critical and important?

chebum (dismissed):
I have fixed unused imports warning and fixed that theme extension using defaultProps actually works.

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.stories.tsx
chebum: I added a corresponding story

PR #257, opened 2024-08-20
fix(DEV-12311): taxes value does not appear

vkuprin:


chebum (changes requested):
As far as I can see, the problem is still present:
Screenshot 2024-08-21 at 11 49 33

PR #260, opened 2024-08-20
refactor(proposal): tune ESLint rules and apply them ❇️

radist2s:
1. **Enabled and Applied `no-extra-boolean-cast` Rule:**
   - Automatically fixed cases where unnecessary boolean casts were present.

2. **Enabled and Applied `prefer-const` Rule:**
   - Ensured that variables which are never reassigned are declared as `const` for better immutability.

3. **Enabled and Manually Applied `@typescript-eslint/no-unused-vars` Rule:**
   - Removed or adjusted variables that were declared but never used, reducing code clutter.


chebum (approved):
Code is easier to read after the change. I approve it.

vkuprin (approved):

PR #261, opened 2024-08-21
fix(DEV-12336): fix table filters layout and make them responsive

chebum:
This PR fixes layout of table filters


packages/sdk-themes/src/themes/monite.ts
```diff
@@ -318,6 +326,26 @@ export const defaultMoniteComponents: Components<Omit<Theme, 'components'>> = {
...
+          width: '100%',
+        },
+      },
+    },
+  },
+  MuiStack: {
+    styleOverrides: {
+      root: {
+        '&.Monite-Filters': {
+          '& > *': {
```
chebum, line 340: @radist2s I made this changes in theme itself, but the filters will look worse in the blank Material UI theme. I can move the layout styles to the sdk itself.

packages/sdk-react/src/components/approvalRequests/ApprovalRequestsTable/ApprovalRequestsFilter/ApprovalRequestsFilter.tsx
```diff
@@ -31,86 +31,78 @@ interface FilterProps {
 export const ApprovalRequestsFilter = ({ onChangeFilter }: FilterProps) => {
   const { i18n } = useLingui();
   const { root } = useRootElements();
+  const className = 'Monite-ApprovalRequestsFilters';
 
   return (
-    <Grid container justifyContent="space-between">
```
radist2s, line 36: The same advice: would be better using just css to pull elements to the right.

packages/sdk-react/src/components/counterparts/CounterpartsTable/Filters/Filters.tsx
```diff
@@ -32,95 +25,89 @@ export const Filters = ({ onChangeFilter, showCategories }: Props) => {
...
+          <InputLabel id="category">{t(i18n)`Category`}</InputLabel>
           <Select
-            labelId="type"
-            label={t(i18n)`Type`}
+            labelId="category"
+            label={t(i18n)`Category`}
             defaultValue="all"
             MenuProps={{ container: root }}
             onChange={(search) => {
-              onChangeFilter(FILTER_TYPE_TYPE, search.target.value);
```
radist2s, line 61: As I see, this refactoring with the selectors `onChangeFilter` and `value`s should be a part for separate PR. Moreover, this refactoring the code without need adds difficulties for other developers on merges.
Btw, I would not change options to `'false' | 'true'`

radist2s (changes requested):

packages/sdk-react/src/components/misc/FilterContainer.tsx
```diff
@@ -0,0 +1,53 @@
+import { ReactNode } from 'react';
+
+import { classNames } from '@/utils/css-utils';
+import { Stack } from '@mui/material';
+
+interface Props {
```
vkuprin, line 6: We have a convention to name them as `ComponentNameProps`, so in this case `FilterContainerProps`

vkuprin (approved):

PR #263, opened 2024-08-21
feat(DEV-12191): improved grid autosize

chebum:
This PR: 
- Reduces amount of layout jumps on page load,
- Preserves column state between page reloads.

### DIFF
--- a/packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
+++ b/packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
@@ -1,0 +1,137 @@
+import { useLayoutEffect, useRef } from 'react';
+import { flushSync } from 'react-dom';
+
+import { useMoniteContext } from '@/core/context/MoniteContext';
+import { GridColDef, useGridApiRef } from '@mui/x-data-grid';
+
+export const defaultCounterpartColumnWidth = 250;
+const maximumDocumentIdColumnWidth = 200;
+
+export function useAreCounterpartsLoading(
+  rows?: { counterpart_id?: string }[]
+) {
+  const { api } = useMoniteContext();
+  const result = !!api.counterparts.getCounterpartsId.useIsFetching({
+    infinite: false,
+    exact: false,
+    predicate: (query) =>
+      !!rows?.some(
+        ({ counterpart_id }) =>
+          !query.state.isInvalidated &&
+          counterpart_id === query.queryKey[1].path.counterpart_id
+      ),
+  });
+  return !rows ? true : result;
+}
+
+// Adapted from https://mui.com/x/react-data-grid/column-dimensions/#autosizing-asynchronously
+// setTimeout and flushSync are necessary for call order control
+// Docs say:
+// The Data Grid can only autosize based on the currently rendered cells.
+// DOM access is required to accurately calculate dimensions
+export function useAutosizeGridColumns(
+  rows: any,
+  columns: GridColDef[],
+  areCounterpartsLoading: boolean,
+  columnSerializationKey: string
+) {
+  const gridApiRef = useGridApiRef();
+
+  const autoSizePerformed = useRef(false); // useRef instead of useState to avoid re-renders on value changes
+  const serializationKey = 'Monite-DataGridColumns-' + columnSerializationKey;
+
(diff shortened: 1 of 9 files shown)


packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
```diff
@@ -0,0 +1,137 @@
...
+      );
+    }
+
+    return () => {
+      timeouts.forEach(window.clearTimeout); // Cancel pending timeouts on unmount
+    };
+  }, [gridApiRef, rows, columns, areCounterpartsLoading, serializationKey]);
+
+  // Save columns width when switching between pages
+  useLayoutEffect(() => {
```
vkuprin, line 112: Consider to add early return for trivial cases ` if (!rows || !gridApiRef.current) return; `

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
```diff
@@ -0,0 +1,137 @@
...
+    const grid = gridApiRef.current;
+    const serializedColumnsStr = localStorage.getItem(serializationKey);
```
vkuprin, line 114: I thin in that case we actually should  avoid excessive local storage access. Consider that to wrap in useMemo
` useMemo(() => localStorage.getItem(serializationKey), [serializationKey]);`

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
```diff
@@ -0,0 +1,137 @@
...
+    const timeouts: number[] = [];
+    if (
+      rows && // Do not autosize until rows are loaded
+      rows.length &&
+      grid &&
+      grid.autosizeColumns && // Will be undefined in test environment
+      grid.getRowsCount() && // Make sure the grid was already rendered so that autosizeColumns function will work
+      grid.getCellElement(rows[0].id, grid.getAllColumns()[0].field) && // Make sure the grid was already rendered so that autosizeColumns function will work
+      !areCounterpartsLoading && // Skip autosize until counterparts are loaded
+      !autoSizePerformed.current && // Do not perform autosize twice
```
vkuprin, line 55: Consider setting `autoSizePerformed.current = true` immediately 

```ts

// Before:
if (!autoSizePerformed.current) {
    // Autosize logic
}

// After:
if (!autoSizePerformed.current && !areCounterpartsLoading) {
    autoSizePerformed.current = true;  // Prevent future executions
    // Autosize logic
}
```

 In that case we avoid the potential for multiple executions of autosizing logic if the effect reruns before the current autosize operation completes

radist2s (approved):
💎

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
chebum: Checking for rows count makes no sense for deserialization of column's width. I added check for columns number though.

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
chebum: We need this value to be sustain page re-loads and component unmounts. useMemo will be cleared on component unmount.

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
```diff
@@ -0,0 +1,137 @@
...
+      !localStorage.getItem(serializationKey) // Do not perform autosize on next page loads - we preserve column state, set by user
+    ) {
+      timeouts.push(
+        window.setTimeout(() => {
+          flushSync(() => {
```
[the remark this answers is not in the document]
chebum: Yes, I tried. flushSync was added as per DataGrid documentation. I added a link to the corresponding doc on top of the file

packages/sdk-react/src/core/hooks/useAutosizeGridColumns.ts
chebum: autoSizePerformed.current is used for determining if current columns state should be saved. Therefore, this value should be set only after actually performing the autosizing operation

PR #270, opened 2024-08-22
feat(DEV-12352): show counterpart avatars

chebum:
This PR adds avatars next to counterpart's name


radist2s:
@chebum, could you investigate the possibility of how to set pictures for avatars? Previously we had API restrictions, and I'm wondering if we can use the avatars of Clerk users to display as avatars.

PR #277, opened 2024-08-26
feat(DEV-11776): colored counterpart logos

chebum:
- This PR adds coloured counterpart logos
- Refactors code to use the same CounterpartCell component for all data grids
- Fixes CounterpartsTable statuses to look the way they were designed in Figma

This PR also includes code from https://github.com/team-monite/monite-sdk/pull/261 - merge it first!


packages/sdk-react/src/components/counterparts/CounterpartStatusChip/CounterpartStatusChip.tsx
```diff
@@ -0,0 +1,104 @@
...
+        case 'size':
+          return true;
+        default:
+          return false;
+      }
+    },
+  }
+)({});
+
+const getLabel = (i18n: I18n, status: 'customer' | 'vendor'): string => {
```
vkuprin, line 95: We ussualy don't pass return type `vendor'): string` as it's not done in entire project. The typescript can identify here himself return type

packages/sdk-react/src/components/counterparts/CounterpartStatusChip/CounterpartStatusChip.tsx
```diff
@@ -0,0 +1,104 @@
...
+  switch (status) {
+    case 'customer':
+      return t(i18n)`Customer`;
+    case 'vendor':
+      return t(i18n)`Vendor`;
+    default:
+      throw new Error(`Unknown counterpart status: ${status}`);
```
vkuprin, line 102: Do we really want to crash component in case if not status was passed, maybe better to return string and console.error?

packages/sdk-themes/src/themes/monite.ts
```diff
@@ -113,6 +118,14 @@ const statusColors: {
   },
 };
 
+const counterpartColors: string[] = [
+  'rgba(0,0,255,0.05)',
+  'rgba(255,0,32,0.05)',
+  'rgba(0,255,220,0.05)',
+  'rgba(225,1,251,0.05)',
+  'rgba(255,123,0,0.05)',
+];
```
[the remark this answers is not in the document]
chebum: I'm using these colors randomly, by index. Color index is calculated based on the customer's name

packages/sdk-react/src/ui/CounterpartCell/CounterpartCell.tsx
```diff
@@ -1,77 +1,69 @@
...
+}: {
+  name: string;
+  isLoading?: boolean;
+}) => {
+  if (!name) return null;
+  const nameParts = name.split(' ');
+  const avatarLetters = (
+    nameParts.length >= 2
+      ? nameParts[0][0] + nameParts[nameParts.length - 1][0]
+      : name[0] || ''
```
vkuprin, line 30: Some of the logic here looks like a magic numbers to me as has not dive deep. What you think to provide some extra comments here or reduce complexity?

Maybe something like:
`nameParts.length >= 2: // Checks if there are at least two parts in the nameParts array.
`
...

packages/sdk-react/src/components/counterparts/CounterpartStatusChip/CounterpartStatusChip.tsx
chebum: Fixed

vkuprin (approved):
Thanks for making couple of adjustments! LGTM

PR #279, opened 2024-08-27
feat(DEV-11776): improve compose email form

chebum:
This PR aims to improve "Compose Email" experience. Adds the following capabilities:
- Open invoice details after its creation,
- Select invoice recipient,
- Preview invoice email before sending.


packages/sdk-react/src/core/queries/useCounterpart.ts
```diff
@@ -296,26 +296,129 @@ export const useDeleteCounterpartVat = (counterpartId: string) => {
...
       path: { counterpart_id: counterpartId ?? '' },
     },
     {
       enabled: Boolean(counterpartId && counterpart?.type === 'organization'),
     }
   );
+
+  if (counterpart?.type === 'individual') {
+    const individual = (
+      counterpart as components['schemas']['CounterpartIndividualRootResponse']
```
vkuprin, line 380: We have type guard functions that allow us to prevent using type assertions. 

Please see :

```ts
export const isIndividualCounterpart = (
  counterpart: CounterpartResponse
): counterpart is components['schemas']['CounterpartIndividualRootResponse'] =>
  counterpart.type === 'individual';
```
`packages/sdk-react/src/components/counterparts/helpers.ts`

packages/sdk-react/src/core/hooks/useFormPersist.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string,
+  getValues: () => { [key: string]: any },
+  setValue: SetFieldValue<any>
+) => {
+  useEffect(() => {
+    const storage = window.sessionStorage;
+    const str = storage.getItem(name);
+
+    if (str) {
+      const values = JSON.parse(str);
```
vkuprin, line 16: For a JSON.parse it's agood practise to wrap it with try catch

packages/sdk-react/src/core/hooks/useFormPersist.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect } from 'react';
+import { SetFieldValue } from 'react-hook-form';
+
+// Persists form state between page reloads
+
+export const useFormPersist = (
+  name: string,
+  getValues: () => { [key: string]: any },
```
vkuprin, line 8: This type can be more specific; if possible, replace `any` with more defined types of your form values.

packages/sdk-react/src/core/hooks/useFormPersist.ts
```diff
@@ -0,0 +1,34 @@
...
+  setValue: SetFieldValue<any>
```
vkuprin, line 9: SetFieldValue<any> is broad; consider narrowing this type to improve type safety

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -520,3 +521,81 @@ export function useInvoiceDetails({
...
+
+  const refresh = () => {
+    setPreview('');
+    setError('');
+    setIsLoading(true);
+    setAttemptNumber(attemptNumber + 1);
+  };
+
+  const language = () => {
+    let locale = i18n.locale;
```
vkuprin, line 567: In this project, we aim to avoid using ‘let’ and instead use ‘const’ for improved readability. The idea, as previously discussed with the team, is to prevent situations where ‘let’ is declared at the top of the file and then mutated further down. Therefore, it’s best to avoid using ‘let’ altogether in this case.

chebum: addressed

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -520,3 +521,81 @@ export function useInvoiceDetails({
...
+          t(i18n)`Failed to update receivable line items: ${errorMessage}`
+        );
+      },
+    }
+  );
+
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [preview, setPreview] = useState<string>('');
+  const [error, setError] = useState<string>('');
+  const [attemptNumber, setAttemptNumber] = useState(0);
```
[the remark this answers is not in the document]
chebum: Thank you for the finding - I addressed the issue.

packages/sdk-react/src/core/hooks/useFormPersist.ts
chebum: Thank you for the suggestion. I updated the code.

packages/sdk-react/src/core/queries/useCounterpart.ts
```diff
@@ -296,26 +297,129 @@ export const useDeleteCounterpartVat = (counterpartId: string) => {
...
       path: { counterpart_id: counterpartId ?? '' },
     },
     {
       enabled: Boolean(counterpartId && counterpart?.type === 'organization'),
     }
   );
+
+  if (counterpart && isIndividualCounterpart(counterpart)) {
+    const individual = (
+      counterpart as components['schemas']['CounterpartIndividualRootResponse']
```
vkuprin, line 381: I think you can remove this type assetion now as we have that wrapped with type guard

vkuprin (approved):

PR #281, opened 2024-08-27
fix(DEV-12180): prevent marking invoices due today as overdue

vkuprin:


packages/sdk-react/src/components/payables/utils/isInvoiceOverdue.ts
```diff
@@ -14,6 +16,7 @@ export const isInvoiceOverdue = (
   return (
     statuses.includes(invoice.status) &&
     invoice.due_date &&
-    new Date(invoice.due_date).getTime() < today.getTime()
+    new Date(invoice.due_date).getTime() < today.getTime() &&
+    new Date(invoice.due_date).getTime() < today.getTime() + 24 * 60 * 60 * 1000
```
chebum, line 20: A comparison a < 10 && a < 20 can be simplified to a < 10. I suggest writing a unit-test to cover all edge-cases. The test will be very simple and it won't take much time to write.

chebum (changes requested):

packages/sdk-react/src/components/payables/utils/isInvoiceOverdue.ts
vkuprin: Dicussed directly with @chebum , agreed to cover with unit test

packages/sdk-react/src/components/payables/PayableDetails/PayableDetails.test.tsx
```diff
@@ -609,5 +611,64 @@ describe('PayableDetails', () => {
...
+        {
+          description: 'should return false when due_date is after today',
+          status: PayableStateEnum.WAITING_TO_BE_PAID,
+          due_date: new Date(Date.now() + 48 * 60 * 60 * 1000).toISOString(),
+          expected: false,
+        },
+        {
+          description: 'should return false when due_date is today',
+          status: PayableStateEnum.WAITING_TO_BE_PAID,
+          due_date: new Date().toISOString(),
```
chebum, line 640: This will return string with current timestamp. Therefore check will work in tests, but will fail in real world.
```
  if (new Date(invoice.due_date).getTime() === new Date().getTime())
    return false;
```
You need test dates without time in them.

chebum (changes requested):

chebum (approved):

PR #287, opened 2024-08-28
feat(DEV-12417):  table empty state not content to display or error message cases

vkuprin:


packages/sdk-react/src/components/counterparts/CounterpartsTable/CounterpartsTable.tsx
```diff
@@ -355,6 +366,46 @@ const CounterpartsTableBase = ({
...
+    return (
+      <DataGridEmptyState
+        icon={
+          <ErrorIcon sx={{ fontSize: '4rem', color: 'error.main', mb: 2 }} />
+        }
+        title={t(i18n)`Failed to Load Counterparts`}
+        descriptionLine1={t(i18n)`There was an error loading counterparts.`}
+        descriptionLine2={t(i18n)`Please try again later.`}
+        actionButtonLabel={t(i18n)`Reload`}
+        onAction={() => window.location.reload()}
```
chebum, line 403: While this approach is fine for <iframe> integration, it may have unexpected consequencies when our components were integrated as react components. I suggest to use useQuery().refetch instead:
`onAction={() => refetch()}`

packages/sdk-react/src/components/counterparts/CounterpartsTable/CounterpartsTable.tsx
```diff
@@ -355,6 +366,46 @@ const CounterpartsTableBase = ({
     return <AccessRestriction />;
   }
 
+  if (!isLoading && counterparts?.data.length === 0) {
+    return (
+      <DataGridEmptyState
+        icon={
+          <NoDataIcon sx={{ fontSize: '4rem', color: 'primary.main', mb: 2 }} />
```
chebum, line 373: This code setting an icon repeats across every table component. Let's move it to the DataGridEmptyState itself and remove the icon property. There is no need to implement customizability "just in case", until it's not strictly required.

chebum (changes requested):

chebum (approved):
Thank you for the implementation of empty states.

PR #305, opened 2024-09-04
feat(DEV-12499): use tax value for us entity during invoice creation

PeCrio:
- Use tax value for US entities during invoice creation.
- Don't require counterpart vat id on create receivable for US based counterparts

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -88,6 +88,11 @@ const CreateReceivablesBase = ({
 
   const createReceivable = useCreateReceivable();
   const { api, monite } = useMoniteContext();
+  const { data: entity, isLoading: isEntityLoading } =
+    api.entityUsers.getEntityUsersMyEntity.useQuery();
+
+  const isUSEntity = entity?.address.country === 'US';
+
   const { data: settings, isLoading: isSettingsLoading } =
     api.entities.getEntitiesIdSettings.useQuery({
       path: { entity_id: monite.entityId },
@@ -111,7 +116,7 @@ const CreateReceivablesBase = ({
 
   const theme = useTheme();
 
-  if (isSettingsLoading) {
+  if (isSettingsLoading || isEntityLoading) {
     return <LoadingPage />;
   }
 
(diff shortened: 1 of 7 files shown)


packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
```diff
@@ -88,6 +88,12 @@ const CreateReceivablesBase = ({
 
   const createReceivable = useCreateReceivable();
   const { api, monite } = useMoniteContext();
+  const { data: entity, isLoading: isEntityLoading } =
+    api.entityUsers.getEntityUsersMyEntity.useQuery();
+
+  // TODO: This can be moved up to a context and shared
+  const isUSEntity = entity?.address.country === 'US';
```
vkuprin, line 95: I suggest you to create a separate utility function and export it in multiple places:

```ts
export const checkIfUSEntity = (entity) => {
  return entity?.address?.country === 'US';
};
```

PeCrio: Thanks @vkuprin. I agree this should be reusable, which is why I added a comment to the usage. Ideally, it would be nice to provide `entity` related values from an umbrella context with this type of unique check done once but we can introduce this later if need be. In the meantime, where you would suggest the utility function be added so it flows with the current folder structure/pattern.

vkuprin: For the Payables module, we have a utils folder where we keep all the small helper pure functions. You can apply the same concept to the Receivables module by creating a utils folder and placing your functions there. Additionally, move the getStatusLabel function to this folder as well, since it’s currently located at the root of the Receivables module and is the only helper or utility function in this module. Now that there are two utility functions, it makes sense to create a utils folder to organize them.

Screenshot 2024-09-04 at 18 26 04

PeCrio: I would reduce the size of the PR increasing, by reducing refactors in moving other utility functions and their use-cases especially because of the time sensitive task. In the meantime, I have added this function to the `countries` utility file since it's a similar context and can be reused more here. You can let me know if you want this to still be moved before the PR is merged.

PR #312, opened 2024-09-12
feat(DEV-12593): Merge create/preview steps of invoice creation into a single flow

chebum:
This PR opens invoice for editing immediately after its creation. Previously the user were directed back to the invoices list.

The PR contains a long test that fills required fields when creating a new Receivable. I didn't mock nested components because we had a broken test that tested Receivable creation. Therefore my longer test both restores the disabled test from 2022 and tests the new receivable creation flow.

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.test.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.test.tsx
@@ -1,17 +1,30 @@
+import { VirtuosoMockContext } from 'react-virtuoso';
+
+import { Receivables } from '@/components';
 import { CounterpartDataTestId } from '@/components/counterparts/Counterpart.types';
-import { getCounterpartName } from '@/components/counterparts/helpers';
 import { CreateCounterpartDialogTestEnum } from '@/components/receivables/InvoiceDetails/CreateReceivable/sections/components/CreateCounterpartDialog.types';
-import { counterpartListFixture } from '@/mocks';
 import { entityIds, entityVatIdList } from '@/mocks/entities';
 import { paymentTermsFixtures } from '@/mocks/paymentTerms';
 import { DateTimeFormatOptions } from '@/utils/DateTimeFormatOptions';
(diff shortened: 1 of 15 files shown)


chebum:
The PR contains a long test that fills required fields when creating a  new Receivable. I didn't mock nested components because we had a broken test that tested Receivable creation. Therefore my longer test both restores the disabled test from 2022 and tests the new receivable creation flow.

packages/sdk-react/src/utils/test-utils.tsx
```diff
@@ -211,6 +212,34 @@ export async function waitUntilTableIsLoaded(
...
+ *
+ * @param predicate Predicate checking for condition
+ * @param timeout Wait function timeout
+ */
+export function waitForCondition(
+  predicate: () => boolean,
+  timeout: number = 1_000
+) {
+  return new Promise<void>((resolve, reject) => {
+    const interval = 50;
```
vkuprin, line 226: Can we have this as passed argument with default value 50?

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.test.tsx
```diff
@@ -44,37 +57,100 @@ describe('CreateReceivables', () => {
...
+      timeout: 3_000,
     });
-  });
+
+    // Create invoice
+    const nextPageButton = await nextPageButtonPromise;
+    fireEvent.click(nextPageButton);
+
+    // Compose email dialog opens
+    await waitForCondition(
```
vkuprin, line 149: Can we please make some cleanups here and remove // comments as code does looks as self explanatory

chebum: I removed comments as suggested

packages/sdk-react/src/utils/test-utils.tsx
chebum: Passed the check interval as a function parameter

vkuprin (approved):

PR #314, opened 2024-09-13
chore(proposal): API version to 2024-05-25

vkuprin:


packages/sdk-react/src/mocks/entities/entitiesFixture.ts
```diff
@@ -11,6 +11,9 @@ function getEntitySettings(): MergedSettingsResponse {
     allow_purchase_order_autolinking: false,
     payment_priority: 'balanced',
     receivable_edit_flow: 'non_compliant',
+    generate_paid_invoice_pdf: faker.datatype.boolean(),
+    quote_signature_required: faker.datatype.boolean(),
+    vat_mode: faker.datatype.boolean() ? 'inclusive' : 'exclusive',
```
chebum, line 16: Random entity settings add a lot of unpredictability to tests. With random settings tests are also harder to debug since data changes between test runs. I recommend to change it to either 'inclusive' or 'exclusive'. Should we need to test behavior of the code with different setting, a different entity should be created for test purposes.

chebum (approved):

PR #316, opened 2024-09-16
feat(DEV-12598): checks for US entity logic cherry-picked

PeCrio:
Checks for US entity logic cherry-picked

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
@@ -353,9 +353,8 @@ export const ItemsSection = ({
                                 defaultValue=""
                                 MenuProps={{ container: root }}
                                 onChange={(e) => {
-                                  const vatRateId = e.target.value;
                                   const vatRate = vatRates?.data.find(
-                                    (vatRate) => vatRate.id === vatRateId
+                                    (vatRate) => vatRate.id === e.target.value
                                   );
 
                                   if (!vatRate) {
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
@@ -329,6 +329,7 @@ export const ItemsSection = ({
                                 InputProps={{
                                   endAdornment: '%',
                                 }}
+                                sx={{ minWidth: 100 }}
                               />
                             </FormControl>
                           )}
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
@@ -74,6 +74,7 @@ function prepareLineItem(
     measure_unit_id: product.measure_unit_id,
     vat_rate_id: vatRates?.data[0].id,
     vat_rate_value: vatRates?.data[0].value,
+    tax_rate_value: 0,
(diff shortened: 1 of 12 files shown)


packages/sdk-react/src/core/queries/useMe.ts
```diff
@@ -0,0 +1,24 @@
...
+  const { api } = useMoniteContext();
+
+  const queryProps = api.entityUsers.getEntityUsersMyEntity.useQuery(
+    {},
+    {
+      retry: false,
+      staleTime: Infinity,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+    }
```
vkuprin, line 13: We usually don’t pass specific configurations for queries. Please refer to other examples, such as in `packages/sdk-react/src/core/queries/useCounterpart.ts`

 Ideally, it should just be:
`const queryProps = api.entityUsers.getEntityUsersMyEntity.useQuery()`

vkuprin (changes requested):
Looks good! Just only one improvement needed

packages/sdk-react/src/core/queries/useMe.ts
PeCrio: That particular section was `cherry-picked`, from the options being passed, I can also understand why because this is a fetch for a user top-level information (entity) which ideally shouldn't change or at least frequently that would require retries, stale time before refetching the data, fetching when focus is shifted in and out of the browser window or mount. 

Except if these options are passed by default, I think it's nice they are added to that specific use-case, but you can advise otherwise why they shouldn't be; I am happy to remove them as well.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
```diff
@@ -300,50 +306,80 @@ export const ItemsSection = ({
...
+                            >
+                              <Select
+                                {...field}
+                                id={field.name}
+                                labelId={field.name}
+                                size="small"
+                                defaultValue=""
+                                MenuProps={{ container: root }}
+                                onChange={(e) => {
+                                  const vatRateId = e.target.value;
```
vkuprin, line 356: I encourage you not to create a new variable if it’s only used once, as it allocates additional memory unnecessarily. Currently, you’re using it only in `(vatRate) => vatRate.id === vatRateId`. Using it directly as `(vatRate) => vatRate.id === e.target.value` is perfectly fine and is a micro-optimization

packages/sdk-react/src/core/queries/useMe.ts
vkuprin: Justifying them for this specific scenario seems reasonable from your explanation. I would just ask then to add extra comment section on of the function explaning that, because we are introducing a deviation from the default settings 

/**
 * @note We are deviating from the default query configuration because the data 
 * does not require frequent refetching or retries as the data fetched
 * (user entity information) is relatively static
 * 
 * @returns {QueryResult} The result of the user entity fetch.
 */
export const useMyEntity = () => {

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
PeCrio: That was a pre-existing code, but I am happy to modify it as I agree with your comment.

packages/sdk-react/src/core/queries/useMe.ts
PeCrio: Comment added.

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/ItemsSection.tsx
PeCrio: Done

vkuprin (approved):

PR #318, opened 2024-09-19
fix(proposal): analytics query to continuosly spawn on every render

vkuprin:


packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx
```diff
@@ -504,7 +504,16 @@ const PayablesTableBase = ({
...
 
-  return api.payables.getPayablesAnalytics.useQuery();
+  return queryClient
+    ? api.payables.getPayablesAnalytics.useQuery(
+        {},
+        {
+          enabled: true,
+        }
+      )
+    : { data: null, isLoading: true };
```
PeCrio, line 518: Did you intend to set the loading state to true when the `queryClient` isn't set?

vkuprin: If queryClient is not ready, we return a default state { data: null, isLoading: true } instead of calling the query

PeCrio (approved):

PR #324, opened 2024-09-20
feat(DEV-12363): move order of columns - extend config to pass order

vkuprin:


packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx
```diff
@@ -202,7 +203,28 @@ const PayablesTableBase = ({
...
+      'counterpart_id',
+      'created_at',
+      'issued_at',
+      'due_date',
+      'status',
+      'amount',
+      'pay',
+    ],
+    []
+  );
```
PeCrio, line 218: Is there a unique use case for utilising `useMemo`  here? I see this is an inexpensive operation that doesn't require optimisation/caching by this hook. 

It is a constant that doesn't run any logic or depend on any variable within the component; it can even be moved outside the component. Depending on the folder structure or reusability, it can be exported/imported externally or just moved outside the component within the same file.

PeCrio (changes requested):

packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx
vkuprin: This useMemo was recommended by eslint

Screenshot 2024-09-20 at 15 45 09

vkuprin: Thanks, completely forgot we can avoid this complexity by moving out as a const

PR #329, opened 2024-09-23
feat(DEV-12590): email invoice details cherry-picked post demo

PeCrio:
- Email invoice details cherry-picked post demo.
- Introduces refactors and cleanups as well.


packages/sdk-react/src/core/queries/useMe.ts
```diff
@@ -29,3 +48,12 @@ export const useMyEntity = () => {
...
 };
+
+export const useMyEntityName = () => {
+  const { data: myEntity } = useMyEntity();
+
+  if (isOrganizationEntity(myEntity)) return myEntity.organization.legal_name;
+  else if (isIndividualEntity(myEntity))
+    return myEntity.individual.first_name + ' ' + myEntity.individual.last_name;
+  return '';
+};
```
vkuprin, line 59:
```suggestion
export const useMyEntityName = () => {
  const { data: myEntity } = useMyEntity();

  if (!myEntity) return '';

  if (isOrganizationEntity(myEntity)) {
    return myEntity.organization.legal_name;
  }

  if (isIndividualEntity(myEntity)) {
    const { first_name, last_name } = myEntity.individual;
    return `${first_name} ${last_name}`;
  }

  return '';
};
```

- Added null check: if (!myEntity) return '' prevents errors if myEntity is undefined or null.
- Improved readability: Destructuring const { first_name, last_name } makes the code cleaner when accessing multiple values.
- Used template literals: Replaced string concatenation with template literals for simpler and more modern syntax.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -72,17 +76,65 @@ const EmailInvoiceDetailsBase = ({
...
+${me.first_name}`;
+        previousBody.current = newBody;
+        setValue('body', newBody, {
+          shouldValidate: false,
+          shouldDirty: true,
+          shouldTouch: true,
+        });
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
vkuprin, line 115: Why we are disalbing here eslint?

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -72,17 +76,65 @@ const EmailInvoiceDetailsBase = ({
...
+  useEffect(() => {
+    if (me && to && contacts && body == previousBody.current) {
+      const contact = contacts.find(({ email }) => email == to);
+      if (contact) {
+        const newBody = t(i18n)`Hi ${contact.first_name},
+
+Please find the invoice attached as discussed.
+
+Kind Regards,
+${me.first_name}`;
```
vkuprin, line 106: Can we please fix here identation problem, I see prettier can't format this easily

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -72,17 +76,65 @@ const EmailInvoiceDetailsBase = ({
...
+        to: '',
+        // TODO: add support for multiple recipients, cc and bcc fields
+      },
+    });
+
+  const { data: me } = useMe();
+  const { data: receivable } = useReceivableById(invoiceId);
+  const { data: contacts } = useReceivableContacts(invoiceId);
+  const to = watch('to');
+  const body = watch('body');
```
vkuprin, line 95: I think you can use oneWatch to listen for multiple values to not create extra varriable

.changeset/pretty-dodos-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+improvements to email invoice functionality
```
saladNights, line 5: add the task number so it will match our guidelines for changeset messages

saladNights (changes requested):

PeCrio (commented):
This pull request is a cherry-picked copy of the code used during demo. I can see a couple of refactors that can be introduced aside the comments added which would improve readability and performance. Currently working on them, e.g. how default values are retrieved for the form.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
PeCrio: This won't be needed in the refactors I am introducing.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
PeCrio: This won't be needed anymore in the refactors I am introducing.

PeCrio (commented):
There are a couple more improvements that can be introduced especially moving out logic/components to separate files. But can also do this as a separate PR improvement due to time.

vkuprin (approved):

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -59,30 +62,95 @@ interface EmailInvoiceDetailsProps {
...
+
+  const to = defaultContact?.email ?? '';
+
+  const body =
+    defaultContact && me
+      ? t(i18n)`Hi ${defaultContact.first_name},
 
-export const EmailInvoiceDetails = (props: EmailInvoiceDetailsProps) => (
-  <MoniteScopedProviders>
-    <EmailInvoiceDetailsBase {...props} />
```
saladNights, line 65: what is the reason to split logic into `EmailInvoiceDetails` & `EmailInvoiceDetailsBase`?
Usually, the pattern is as follows:
we put everything the component needs into the `ComponentBase`, then wrap it with `MoniteScopedProviders` and import as the `Component`.
Moreover, we don't have to wrap those components with `MoniteScopedProviders` that we don't export from the SDK. Could you consider removing this wrapper? Probably one of its parents already wrapped with it.

saladNights (changes requested):

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
PeCrio: Thanks for that context. The split (`EmailInvoiceDetails` & `EmailInvoiceDetailsBase`) and `MoniteScopedProviders` are part of already existing code before the cherry-picking, but I am happy to make further modifications.

I would want to keep the pattern of passing down the default form values as props for the `Form` compared to the previous approach as it extends reusability for other use cases; it's also a recommended approach for the react-hook-form library for resetting form values, improves performance, and reduces the number of re-renders and potential of unexpected side effects from the checks in the previous implementation.

saladNights (approved):

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
saladNights: ok, let's leave it then.
I agree with approach of passing down the default form values

PR #340, opened 2024-09-25
fix(proposal): custom style cards enhanced

vkuprin:


.changeset/two-kings-study.md
```diff
@@ -0,0 +1,2 @@
+---
```
PeCrio, line 1: Changeset seems empty.

vkuprin: As I know, we don't put changeset info, if it's small addition

saladNights (approved):

PR #387, opened 2024-10-28
feat(DEV-12623): manual payment record implementation

PeCrio:


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/TabPanels/PaymentTabPanel/ManualPaymentRecordDetails.tsx
```diff
@@ -0,0 +1,123 @@
...
+  //   paymentRecords.payment_date ?? ''
+  // ).setHours(
+  //   paymentRecords.payment_time?.getHours() ?? 0,
+  //   paymentRecords.payment_time?.getMinutes() ?? 0
+  // );
+
+  // const dateTime = i18n.date(
+  //   new Date(dateTimeWithReplacedTime),
+  //   dateTimeFormat
+  // );
```
marcperezmonite, line 40: Is this intentional?

PeCrio: Yes, left for a future iteration. Following our earlier conversations on planning on a strategy for managing timezones across the app.

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -590,3 +590,31 @@ export const useReceivableContacts = (receivable_id: string) => {
...
+    {
+      path: {
+        receivable_id,
+      },
+    },
+    {
+      onSuccess: async () => {
+        await Promise.all([
+          api.receivables.getReceivables.invalidateQueries(queryClient),
+          api.receivables.getReceivablesId.invalidateQueries(
```
vkuprin, line 608: I believe we don’t need double invalidation here in the onSuccess handler. Currently, both api.receivables.getReceivables and api.receivables.getReceivablesId are being invalidated. Since marking a specific invoice as paid (by receivable_id) updates the status of that single invoice, invalidating api.receivables.getReceivablesId should be sufficient

packages/sdk-react/src/core/queries/useReceivables.ts
```diff
@@ -590,3 +590,31 @@ export const useReceivableContacts = (receivable_id: string) => {
...
+  const { i18n } = useLingui();
+  const { api, queryClient } = useMoniteContext();
+
+  return api.receivables.postReceivablesIdMarkAsPaid.useMutation(
+    {
+      path: {
+        receivable_id,
+      },
+    },
+    {
```
vkuprin, line 604: It would be helpful to add error handling within the onError callback to notify users of any issues during the mutation

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/TabPanels/PaymentTabPanel/PaymentRecordForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+    event: React.FormEvent<HTMLFormElement>
+  ) => {
+    event.preventDefault();
+
+    handleSubmit(onSubmit)(event);
+  };
+
+  return (
+    <>
+      <form onSubmit={handlePaymentRecordSubmition}>
```
vkuprin, line 74: Could we avoid creating a separate function that only wraps another function? Defining handlePaymentRecordSubmition here simply creates an extra layer of indirection without adding any real value. Instead, it would be more efficient to pass handleSubmit(onSubmit) directly as the onSubmit handler in the form, reducing unnecessary wrapping

vkuprin (changes requested):
Please also remove the excessive comments. If you’d like to keep them for future reference, consider using git stash for temporary notes

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/TabPanels/PaymentTabPanel/UpdatePDF/UpdatePDFSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useLingui } from '@lingui/react';
+import { Box, Typography } from '@mui/material';
+
+import { UpdatePDFModal } from './UpdatePDFModal';
+
+type Props = {
+  invoice: components['schemas']['InvoiceResponsePayload'];
+};
+
+export const UpdatePDFSection: React.FC<Props> = ({ invoice }) => {
```
vkuprin, line 12: We don't use here React.FC, we pass props not as generics but as like this 
`export const UpdatePDFSection= ({ invoice }: Props)`

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/TabPanels/PaymentTabPanel/UpdatePDF/UpdatePDFSection.tsx
```diff
@@ -0,0 +1,46 @@
+import { components } from '@/api';
+import { t } from '@lingui/macro';
+import { useLingui } from '@lingui/react';
+import { Box, Typography } from '@mui/material';
+
+import { UpdatePDFModal } from './UpdatePDFModal';
+
+type Props = {
+  invoice: components['schemas']['InvoiceResponsePayload'];
+};
```
vkuprin, line 10: Please avoid creating a separate type if it’s only used once, as it adds an unnecessary wrapper. Use components['schemas']['InvoiceResponsePayload'] directly instead

packages/sdk-react/src/core/queries/usePaymentIntent.ts
```diff
@@ -0,0 +1,11 @@
+import { useMoniteContext } from '../context/MoniteContext';
+
+export const usePaymentIntentById = (payment_intent_id: string) => {
```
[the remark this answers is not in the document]
PeCrio: I agree with the logic of if a function is only used once it should be associated with the use-case, but for this use-case, this can easily introduce repitition when we need this API handling function for another component, shouldn't we introduce all related API functions like this in the current structure for easier reference and update in the future?

packages/sdk-react/src/components/RHF/RHFTimePicker/RHFTimePicker.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  helperText: isInvalid && error?.message,
+                },
+              }}
+            />
+            {isErrorCustom(error) && (
+              <Alert
+                severity="error"
+                icon={false}
+                sx={{
+                  marginTop: -2,
```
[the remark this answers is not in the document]
PeCrio: This is following the existing pattern currently used by the DatePicker component since they are similar and it seems to be used to have the expected spacing that conforms to design. See `RHFDatePicker.tsx`. You can also advise otherwise.

packages/sdk-react/src/core/queries/useReceivables.ts
PeCrio: I agree, good catch.

packages/sdk-react/src/core/queries/useReceivables.ts
PeCrio: I agree, good catch.

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/TabPanels/PaymentTabPanel/PaymentRecordForm.tsx
PeCrio: I agree.

vkuprin (approved):

PR #427, opened 2024-11-18
Incorrect tax value in generated invoice

arthemium:
Fixed and added * 100 to the tax value as represented in the API docs

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
@@ -164,7 +164,7 @@ const CreateReceivablesBase = ({
           quantity: item.quantity,
           product_id: item.product_id,
           ...(isNonVatSupported
-            ? { tax_rate_value: item?.tax_rate_value ?? 0 * 100 }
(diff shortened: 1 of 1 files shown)


vkuprin (approved):

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/CreateReceivables.tsx
```diff
@@ -164,7 +164,7 @@ const CreateReceivablesBase = ({
           quantity: item.quantity,
           product_id: item.product_id,
           ...(isNonVatSupported
-            ? { tax_rate_value: item?.tax_rate_value ?? 0 * 100 }
+            ? { tax_rate_value: item?.tax_rate_value * 100 ?? 0 * 100 }
```
saladNights, line 167: `item?.tax_rate_value` could be `undefined`, so the current one-line check should be refactored to `tax_rate_value: item?.tax_rate_value ? item?.tax_rate_value * 100 : 0 * 100`. There is also an improvement: 0 * 100 always produces 0, so it could be simplified to` tax_rate_value: item?.tax_rate_value ? item?.tax_rate_value * 100 : 0`

saladNights (changes requested):

PR #436, opened 2024-11-21
chore(proposal): payment improvements for specific flow

vkuprin:


packages/sdk-react/src/components/payables/PayablesTable/hooks/usePaymentHandler.tsx
```diff
@@ -51,20 +41,61 @@ export const usePaymentHandler = (
...
+  const {
+    executeWithRetry: executeWithRetryMarkPaid,
+    isRetrying: isRetryingMarkPaid,
+  } = useRetry({
+    maxAttempts: 5,
+    initialDelayMs: 1000,
+    backoffFactor: 1.5,
+    shouldRetry: () => {
+      if (!payableId) return true;
+      return true;
```
marcperezmonite, line 53: Always true, is this intentional?

vkuprin: This was a partial implementation where I intended to add different behavior based on payableId but this should be aligned based on manual creation flow

marcperezmonite (approved):

PR #445, opened 2024-11-22
feat(proposal): Implement CashCard component for dashboard

oneanotheruser:
-  Implementation CashCard component for demo dashboard
-  Adding header and tabs to demo page

### DIFF
--- a/examples/with-nextjs-and-clerk-auth/src/components/DashboardCard.tsx
+++ b/examples/with-nextjs-and-clerk-auth/src/components/DashboardCard.tsx
@@ -5,7 +5,7 @@ import { Card, CardHeader, CardContent, SvgIconProps } from '@mui/material';
 type IconVariant = 'info' | 'success' | 'critical';
 
 interface DashboardCardProps {
-  title: string;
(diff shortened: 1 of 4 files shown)


examples/with-nextjs-and-clerk-auth/src/components/DashboardCard.tsx
```diff
@@ -5,10 +5,11 @@ import { Card, CardHeader, CardContent, SvgIconProps } from '@mui/material';
 type IconVariant = 'info' | 'success' | 'critical';
 
 interface DashboardCardProps {
-  title: string;
+  title: string | ReactNode;
```
vkuprin, line 8: The ReactNode type in React already includes string as one of its possible types, as shown in your screenshot. According to the ReactNode definition:

Screenshot 2024-11-22 at 12 22 49

This means that ReactNode can already accept string values, so explicitly declaring title: string | ReactNode is redundant

oneanotheruser: Thanks! Updated

PR #446, opened 2024-11-22
feat(proposal): add counterpart bank account holder name

vkuprin:
Added new field bank account holder name for counterpart

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "dev:sdk-demo": "turbo run dev --filter=@team-monite/sdk-demo",
     "build": "turbo run build",
     "clean": "rm -rf packages/*/dist",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
     "dev:sdk-demo": "turbo run dev --filter=@team-monite/sdk-demo",
     "build": "turbo run build",
     "clean": "rm -rf packages/*/dist",
+    "clean:node": "find . -name node_modules -type d -prune -exec rm -rf '{}' + && find . -name .turbo -type d -prune -exec rm -rf '{}' + && find . -name dist -type d -prune -exec rm -rf '{}' +",
```
marcperezmonite, line 21: This doesn't seem related to this task, could you remove it?

vkuprin: Yes, I can keep it in separate PR

marcperezmonite: Better to keep them separated so we can properly track those later.

marcperezmonite (approved):

PR #447, opened 2024-11-22
fix(proposal): set the default text color for tables

saladNights:


packages/sdk-react/src/ui/DueDateCell/DueDateCell.tsx
```diff
@@ -29,9 +29,7 @@ export const DueDateCell = ({ data }: DueDateCellProps) => {
       height="inherit"
       justifyContent="center"
     >
-      <Typography variant="body2" color={overdueDays > 0 ? 'error' : ''}>
-        {formattedDate}
-      </Typography>
+      <span color={overdueDays > 0 ? 'error' : ''}>{formattedDate}</span>
```
vkuprin, line 32: I’d like to retain the <Typography> approach, you can override or extend its styles in the configuration or inline. For example:

```tsx
<Typography 
  variant="body2" 
  component="span"
  color={overdueDays > 0 ? 'error' : 'text.primary'}
>
  {formattedDate}
</Typography>
```

PR #449, opened 2024-11-22
feat(DEV-13064): Prevent click on tables when selecting text

marcperezmonite:


packages/sdk-react/src/utils/text-selection.ts
```diff
@@ -0,0 +1,4 @@
+export function hasSelectedText(): boolean {
+  const selection = window.getSelection();
+  return !!selection && !!selection.toString();
+}
```
vkuprin, line 4: I suggest simplifying it further into an even more concise shorthand

Please consider the following:
```ts
export const hasSelectedText = () =>
  !!window.getSelection()?.toString();
```

marcperezmonite: Thanks for the suggestion. Your suggestion might make the code a bit harder to understand though, do you see any benefit besides style preference?

vkuprin: > `!!` operators are trying to convert values to booleans, but they're not really needed here because:
> 
> - Selection will either be a Selection object or null - when used in boolean context (like in &&), null automatically becomes false
> - Selection.toString() returns a string - empty strings are automatically false in boolean contexts, non-empty strings are true
>
> So we could actually write it as:
> ```ts
> export function hasSelectedText() {
>     const selection = window.getSelection();
>     return selection && selection.toString();
> }
> ```
> 
> Or even more concise with optional chaining:
> ```ts
> export const hasSelectedText = () => window.getSelection()?.toString();
> ```
>
> P.S *Neither is strictly "wrong" - it's about finding the sweet spot between readability and conciseness*

marcperezmonite: Thanks Vitaly. Wouldn't it be the case that neither of those two suggestions would guarantee a boolean return value (they'd change this function to return a string or undefined)?

vkuprin (approved):

packages/sdk-react/src/utils/text-selection.ts
marcperezmonite: One last comment on your last suggestion: `window` could potentially be undefined, so we would still need to check for both. I'll resolve this discussion since your suggestion does not cover fully the current behavior. Thanks for suggesting though!

vkuprin: But we’re already using window in many other places in the app without checking if it’s undefined. I believe we only expect this code to run on the client side, not in an SSR scenario:

Here is our all other cases where we use window:
Screenshot 2024-11-25 at 14 38 31

PR #454, opened 2024-11-27
fix(DEV-13228): Design bugs BillPay

marcperezmonite:


packages/sdk-react/src/components/payables/PayablesTable/Filters/Filters.tsx
```diff
@@ -17,7 +17,6 @@ import { Theme } from 'mui-styles';
 
 import { getRowToStatusTextMap } from '../../consts';
 import {
-  FILTER_TYPE_CREATED_AT,
```
saladNights, line 20: Since you are removing this filter type, this constant could be removed from constants, filter type, and the table data request
`packages/sdk-react/src/components/payables/PayablesTable/consts.ts`
`packages/sdk-react/src/components/payables/PayablesTable/types.ts`
`packages/sdk-react/src/components/payables/PayablesTable/PayablesTable.tsx`

saladNights (changes requested):

packages/sdk-react/src/components/payables/PayablesTable/Filters/Filters.tsx
marcperezmonite: Good catch, let me adjust that

saladNights (approved):

PR #462, opened 2024-12-02
feat(DEV-13257): add min height to table components

saladNights:


PeCrio (commented):
Is there a reason why we are using absolute numbers rather than pixel values?

saladNights:
> Is there a reason why we are using absolute numbers rather than pixel values?

it’s just a shorthand for 500px. But let's make it more explicit

marcperezmonite (approved):

PR #475, opened 2024-12-06
feat(DEV-13301): finance an invoice

PeCrio:


packages/sdk-react/src/components/receivables/Financing/FinanceInvoice/FinanceInvoice.stories.tsx
```diff
@@ -0,0 +1,35 @@
+import { MemoryRouter } from 'react-router-dom';
```
saladNights, line 1: What functionality of the MemoryRouter do you use exactly in the FinanceInvoiceComponent? I can't find any usage cases, and it looks like you can remove it.

saladNights (approved):

PR #492, opened 2024-12-13
fix(proposal): edit AR invoice update button unresponsive

arthemium:
Issue was presented by a customer where the update button in the invoice editing screen was not responsive when the entity is not VAT obligated

Issue:
* tax rate value was not included, hence when VAT value was passed it didn't inculde any values

* note for QA and testing - there's a quirk, when launching the SDK for the first time, the VAT rate IDs will not apply, however, when closing and opening the editing component will reflect the VAT rates and the issue doesn't appear for any other invoice

### DIFF
--- a/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
+++ b/packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
@@ -148,8 +148,6 @@ const EditInvoiceDetailsContent = ({
 
   const formName = `Monite-Form-receivablesDetailsForm-${useId()}`;
 
-  console.log(formName);
-
   const {
     createReminderDialog,
     editReminderDialog,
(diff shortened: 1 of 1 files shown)


saladNights:
update branch with the latest changes form the base one

packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EditInvoiceDetails.tsx
```diff
@@ -99,8 +99,11 @@ const EditInvoiceDetailsContent = ({
       line_items: invoice.line_items.map((lineItem) => ({
         quantity: lineItem.quantity,
         product_id: lineItem.product.id,
-        vat_rate_id: lineItem.product.vat_rate.id,
-        vat_rate_value: lineItem.product.vat_rate.value,
+        ...(lineItem.product.vat_rate.id !== null
+          ? { vat_rate_id: lineItem.product.vat_rate.id,
+              vat_rate_value: lineItem.product.vat_rate.value,
+           }
+          : { tax_rate_value: lineItem.product.vat_rate.value / 100 }),
```
saladNights, line 106: what does the '/ 100' means? If it's the conversion from to minor units, it is a wrong way to do so.
The conversion from/to minor units depends on the currency since there is a differect decimal for some currencies.
You should use formatFromMinorUnits/formatToMinorUnits functions form our util hooks.

saladNights (changes requested):

saladNights:
add a changeset and describe the implemented changes. It is required for detailed release notes

PR #529, opened 2025-01-21
fix(DEV-13770): background styles on Drop-in

marcperezmonite:


packages/sdk-react/src/core/theme/mui-monite/index.ts
```diff
@@ -402,17 +402,9 @@ export const getTheme = (theme: ThemeConfig): ThemeOptions => {
       styleOverrides: {
         root: {
           '&.Monite-Filters': {
-            backgroundColor: 'white',
-            padding: '14px',
+            backgroundColor: moniteTheme.colors.background,
+            padding: moniteTheme.spacing * 1.8,
```
PeCrio, line 406: I will suggest naming "magic numbers" as a CONST for maintainability.

PeCrio (approved):

PR #591, opened 2025-02-20
fix(DEV-12823): manage measure units

ayubova:


marcperezmonite:
@ayubova please run `yarn changesets` and push the new commit. (this should be a minor change for SDK/Drop-in).

marcperezmonite (approved):

PR #593, opened 2025-02-21
fix(DEV-14111): Invoice email send button

costa-monite:
### Main Changes
   - The main change is that the "Send invoice" button now changes its caption based on the invoice status
   - Utility function isInvoiceIssued that checks if an invoice status indicates it has been issued
The function considers an invoice "issued" if its status is one of: 'issued', 'partially_paid', 'paid', or 'overdue'

### Tests
   - `invoiceStatus.test.ts`: Unit tests for the new helper function

Note: The full test suite is currently difficult to implement due to some issues with the fixtures. But for now, helper tests are enough to check the core changes and document them


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/EmailInvoiceDetails.tsx
```diff
@@ -324,7 +324,7 @@ const EmailInvoiceDetailsBase = ({
                   type="submit"
                   form={formName}
                   disabled={isDisabled || isLoading}
-                >{t(i18n)`Issue and send`}</Button>
+                >{t(i18n)`Send`}</Button>
```
marcperezmonite, line 327: Are we sure this text should change always? Or only under certain circumstances?

costa-monite: I have updated the code by adding a helper function that determines if an invoice has been issued

caiomachado-monite (approved):
LGTM

PR #594, opened 2025-02-21
fix(DEV-14020): Custom currency dropdown renderer

costa-monite:
Improved Currency Display:
– When a currency name is too long to fit in the input field, it's now properly truncated with an ellipsis (...) instead of being cut off
– The currency symbol (e.g., $, €, £) remains visible at all times, even when the currency name is truncated
For example, "United States Dollar" might show as "United States Dol... $"

Tooltip:
– When hovering over a truncated currency name, a tooltip appears showing the full currency name and its symbol
Eg, hovering over "United States Dol...$" will show "United States Dollar, $"
This helps users see the complete currency name without needing to make the input field wider

### DIFF
--- a/.changeset/lazy-rabbits-obey.md
+++ b/.changeset/lazy-rabbits-obey.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+Product creation default values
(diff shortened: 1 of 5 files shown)


marcperezmonite:
@costa-monite Patch changesets please 🙏

marcperezmonite (approved):

PR #596, opened 2025-02-24
fix(DEV-14053): vat id error message from api

costa-monite:
- Added a new optional defaultMessage parameter
- Changed the fallback message to use the provided defaultMessage if available, otherwise falls back to Unknown
- The function now returns the custom default message instead of just "Unknown" when no specific error message is found
- Updated the error handling in useCreateCounterpartVat hook
- Instead of showing a generic "Failed to create VAT" message, it now uses getAPIErrorMessage
- Passes the error object and the default message Failed to create VAT to get a more specific error message when available


costa-monite:
- Added a new optional defaultMessage parameter
- Changed the fallback message to use the provided defaultMessage if available, otherwise falls back to Unknown
- The function now returns the custom default message instead of just "Unknown" when no specific error message is found
- Updated the error handling in useCreateCounterpartVat hook
- Instead of showing a generic "Failed to create VAT" message, it now uses getAPIErrorMessage
- Passes the error object and the default message Failed to create VAT to get a more specific error message when available

marcperezmonite (approved):

PR #606, opened 2025-02-25
[Draft] test v3 deployment

costa-monite:


marcperezmonite:
Can we close this @costa-monite ?

PR #609, opened 2025-02-25
fix(proposal): Improve Drop-in components styles to take 100% of available space

marcperezmonite:


costa-monite (approved):
All good, only a rebase on Main is needed

PR #621, opened 2025-02-27
fix(DEV-14232): Prepend https:// as a default value in Onboarding URL field

marcperezmonite:


costa-monite (approved):
LGTM

Unrelated lint issue "Check failure on line 20 in packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/components/InvoicePreview.ts"
might be fixed here as well

PR #628, opened 2025-03-05
fix(DEV-14163): using the organization's email address as a fallback if no contact person is defined

costa-monite:
- Updated the email recipient selector to show organization email when no contacts exist
- Modified email body template to handle both personal and organizational recipients by using the company's name as the default first name


packages/sdk-react/src/components/receivables/InvoiceDetails/ExistingInvoiceDetails/components/helpers/contacts.ts
```diff
@@ -0,0 +1,62 @@
+import { components } from '@/api';
+import {
+  isOrganizationCounterpart,
+  getCounterpartName,
+} from '@/components/counterparts/helpers';
+import { GenericCounterpartContact } from '@/core/queries';
+
+type CounterpartOrganizationRootResponse =
```
caiomachado-monite, line 8: Maybe export this type from here so that you can remove the duplication on the previous file at line 64

costa-monite: good point! thanks
missed that opportunity, but will fix it in the next commit

marcperezmonite:
@costa-monite - Small linting issue + translations, code looks good!

caiomachado-monite (approved):
LGTM

PR #630, opened 2025-03-07
fix(proposal): Dropin new creation syntax

marcperezmonite:
## Drop-in new creation syntax

```ts
import('@monite/sdk-drop-in').then(({ MoniteDropin, MoniteEventTypes }) => {
      const dropinConfig = {
        entityId,
        apiUrl: '...',
        locale: {},
        theme: {},
        componentSettings: {},
        fetchToken: async function fetchToken() {...},
      };

      const dropin = new MoniteDropin(dropinConfig);

      dropin.create(component).mount(moniteAppRef.current);

      dropin
        .on(MoniteEventTypes.INVOICE_CREATED, () => {
          console.log('invoice.created');
        });
});
```

Note: Previous approach is still supported

## Drop-in types
Types are now built and included in the final bundle (see `dist/types`).

### DIFF
--- a/packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
+++ b/packages/sdk-drop-in/src/lib/useIframeAppSlots.tsx
@@ -3,20 +3,18 @@ import { ComponentProps, useCallback, useEffect, useState } from 'react';
 import { MoniteIframeAppCommunicator } from '@/lib/MoniteIframeAppCommunicator';
 import { type APISchema, type MoniteProvider } from '@monite/sdk-react';
 
+type MoniteProviderProps = ComponentProps<typeof MoniteProvider>;
+
 type MoniteProviderSlots = {
-  locale?: ComponentProps<typeof MoniteProvider>['locale'];
-  theme?: ComponentProps<typeof MoniteProvider>['theme'];
-  componentSettings?: ComponentProps<
-    typeof MoniteProvider
-  >['componentSettings'];
+  locale?: MoniteProviderProps['locale'];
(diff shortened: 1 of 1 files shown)


packages/sdk-drop-in/src/custom-elements/monite-app.ts
```diff
@@ -9,9 +14,116 @@ import { MoniteAppElement } from './MoniteAppElement';
...
+      callback,
+      this.moniteAppElement
+    );
+    return this;
+  }
+
+  public off(event: MoniteEventTypes) {
+    this.events[event]?.();
+    return this;
+  }
```
costa-monite, line 125: Looks like we might need a kind of unmount/destroy function with removeChild, clearSlotElements and events cleanup calls. Though this could be implemented later

packages/sdk-drop-in/src/custom-elements/monite-app.ts
```diff
@@ -9,9 +14,116 @@ import { MoniteAppElement } from './MoniteAppElement';
...
+
+  public mount(container: HTMLElement) {
+    container.appendChild(this.moniteAppElement);
+    return this;
+  }
+
+  public on(event: MoniteEventTypes, callback: (event: CustomEvent) => void) {
+    this.events[event] = addMoniteEventListener(
```
costa-monite, line 114: Just to note, we can later identify our event and callback pairs using shared Symbols
something like this: `Symbol.for(`monite_event_${String(event)}_${callback.toString()}`);`

costa-monite (approved):
LGTM (usual build, lint and changeset stuff remained)

PR #635, opened 2025-03-10
feat - DEV-13042 Payables discount implementation

caiomachado-monite:
**Feature overview:**
Discounts has been added to payables API but hadn't been added yet to the SDK, so this PR is to add it.

**Changes Made**
Updated `helper.ts` functions to reflect the newly added discount option

Updated UI on both `PayableDetailsInfo` and `PayableDetailsForm` to display discount and also the ability to add and remove a discount from an invoice


packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -153,6 +158,7 @@ export const prepareSubmit = ({
   counterpartAddressId,
 }: SubmitPayload): components['schemas']['PayableUpdateSchema'] => ({
   document_id: invoiceNumber,
+  discount: discount ?? 0,
```
costa-monite, line 161: The api expects the discount to be a number, but on the client it is nullable
What about initializing it with 0 at the start? Are there any downsides to it?

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
```diff
@@ -207,7 +214,7 @@ export const calculateTotalsForPayable = (
   return {
     subtotal,
     taxes,
-    total,
+    total: discount && discount > 0 ? total - discount : total,
```
costa-monite, line 217: we can safely omit the "truthy" check here `discount > 0 ? total - discount : total`

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
```diff
@@ -145,6 +149,7 @@ const getValidationSchema = (i18n: I18n) =>
             ? schema.required()
             : schema
         ),
+      discount: yup.number().nullable().positive().min(0),
```
costa-monite, line 152: here poisitive() seems redundant as we also have min(0)

costa-monite (approved):
All good, just a couple of suggestions from my side
Also, please check the PR title job

As for the future, in the current state PayableDetailsFormBase handles too many responsibilities: rendering, validation, calculation, and state management
We might consider splitting it into cross cutting components, HOCs etc

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
caiomachado-monite: @costa-monite yeah, the thing is that initially discount comes as null from the API and the goal is that if you want to remove the discount you would send null in the payload, however, this removal of the discount is not working with null and I was asked to send 0 for now, so maybe in the future we can change this but null is definitely needed here

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
caiomachado-monite: @costa-monite Yeah, great point, I had to add min(0) though because otherwise the save button was not allowing the 0 as a valid payload and was blocking the form submit, let me try with just min(0) and then I will remove positive()

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
caiomachado-monite: @costa-monite I tried the first suggestion when I was writing this but since discount can be undefined it was not allowing me to check directly whether it was more than 0 or not, so I had to add the first check, but I believe the second suggestion will work fine

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/PayableDetailsForm.tsx
caiomachado-monite: @costa-monite done

packages/sdk-react/src/components/payables/PayableDetails/PayableDetailsForm/helpers.ts
caiomachado-monite: @costa-monite done

dinis-monite (approved):

PR #647, opened 2025-03-17
feat(DEV-12357): Added ability to hide customer types section from AP/AR/Counterparts

caiomachado-monite:
**Feature overview:**
Customer type section should be customizable whether the user only wants the customer option, the vendor option or both, and if only one option is provided, the section should be hidden altogether.

**Changes Made**
- Added a prop called customerTypes to the componentSettings to allow the users to customize the components through it directly
- Also added the same prop to all the components that will use it directly so that the user can also pass it directly to the component to only change in one place instead
- Did a refactor on a few places, restructured the folder for `counterparts` to remove a bunch of nested folders that were creating confusion, this is still WIP by the way and I will be updating it as we go along
- I also updated all three places to use the new Modal approach for creating a counterpart

**Remarks**
- There is still work to be done in regards to these changes, for example, the view details of a counter part is still using the previous drawer approach, that will also be updated in the future, but for now we are to keep it as is until we receive further instructions
- A lot of changes were in fact just moving a file to another folder so it may say as if I created the whole file but I didn't change it at all


packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartDetails.tsx
```diff
@@ -62,18 +64,25 @@ const CounterpartDetailsBase = (props: CounterpartsDetailsProps) => {
...
         return (
           <CounterpartView
             id={counterpartId}
-            showCategories={showCategories}
+            showCategories={Boolean(
+              (customerTypes && customerTypes?.length > 1) ||
+                (!customerTypes &&
+                  componentSettings?.counterparts?.customerTypes &&
+                  componentSettings?.counterparts?.customerTypes?.length > 1)
+            )}
```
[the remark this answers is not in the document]
caiomachado-monite: @dinis-monite good point Dinis, I will create a variable for this

packages/sdk-react/src/components/counterparts/components/CounterpartAutocomplete.tsx
```diff
@@ -63,24 +73,18 @@ export const CounterpartAutocomplete = <TFieldValues extends FieldValues>({
...
-    setIsCreateCounterpartOpened(true);
-  }, [setIsCreateCounterpartOpened]);
-
-  const handleCloseCreateCounterpart = useCallback(() => {
-    setIsCreateCounterpartOpened(false);
-  }, [setIsCreateCounterpartOpened]);
-
   const { data: counterparts, isLoading: isCounterpartsLoading } =
-    useCounterpartList();
+    useCounterpartList({ query: { is_vendor: true } });
```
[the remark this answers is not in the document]
caiomachado-monite: @dinis-monite Yes, in this component yes because this is only used in Payables and the input literally says "Vendor" in the label haha, so Marc told me to filter this out only to vendors and the same happens in receivables, but we filter only for customers, in the future, we will create only one component that will be used in both places and then we will make it dynamic based on the customerType but for now they asked me to keep it this way

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/components/CounterpartSelector.tsx
```diff
@@ -83,7 +83,7 @@ export const CounterpartSelector = ({
   const { root } = useRootElements();
   const { control, watch } = useFormContext<CreateReceivablesFormProps>();
   const { data: counterparts, isLoading: isCounterpartsLoading } =
-    useCounterpartList();
+    useCounterpartList({ query: { is_customer: true } });
```
costa-monite, line 86: could we safely assume that the counterpart is a customer here?

packages/sdk-react/src/components/counterparts/components/CreateCounterpartModal.tsx
```diff
@@ -59,31 +64,32 @@ export const CreateCounterpartModal = ({
...
-              onCreate={(counterpartId) => {
-                onCreate(counterpartId);
-                onClose();
-              }}
-            />
-          </Grid>
+            defaultValuesOCR={defaultValuesOCR}
+            defaultValues={{
+              isCustomer: customerTypes?.includes('customer'),
+              isVendor: customerTypes?.includes('vendor'),
```
costa-monite, line 84: won't be there a case where neither type is included, might we better assign a customer type here or somewhere upstream?

packages/sdk-react/src/components/receivables/InvoiceDetails/CreateReceivable/sections/components/CounterpartSelector.tsx
caiomachado-monite: @costa-monite yeah, the input label even says: Customer

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartDetails.tsx
```diff
@@ -62,18 +64,27 @@ const CounterpartDetailsBase = (props: CounterpartsDetailsProps) => {
...
-  const { showCategories = true, showBankAccounts = true } = props;
+  const { showBankAccounts = true, customerTypes } = props;
   const defaultValues = props.type ? props.defaultValues : undefined;
 
+  const shouldShowCategories = Boolean(
+    (customerTypes && customerTypes?.length > 1) ||
+      (!customerTypes &&
+        componentSettings?.counterparts?.customerTypes &&
+        componentSettings?.counterparts?.customerTypes?.length > 1)
+  )
```
costa-monite, line 75: could we simplify this a bit:
```typescript
const shouldShowCategories = Boolean(
  (customerTypes?.length ?? 0) > 1 || 
  (componentSettings?.counterparts?.customerTypes?.length ?? 0) > 1
);
```

what do you think?

packages/sdk-react/src/components/counterparts/CounterpartDetails/useCounterpartDetails.tsx
```diff
@@ -85,6 +86,14 @@ interface CommonCounterpartDetailsProps
 }
 
 export type CounterpartsDetailsProps = {
+  /**
+   * Array of available customer types, an array that should contain either customer, vendor, or both.
+   * This array can't be empty and if only one option is passed, the customer type section will be hidden
+   * and the default customer type will be the one passed.
+   * It is set to undefined at component level but defaults to ['customer', 'vendor'] through componentSettings
+   * @param customerTypes - Array of customer types, defaults to ['customer', 'vendor'] through componentSettings
+   */
```
costa-monite, line 95: This comment is repeated like 9 times. Could we consider using some kind of jsdoc tag like `@see`
to define it once and then use it as a link in other places?

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartDetails.tsx
caiomachado-monite: @costa-monite Good idea, although I do need to leave the `!customerTypes` in the second check because that way if the user passes a specific customerTypes array to the component directly, he will be able to override the componentSettings value without affecting the entire application

packages/sdk-react/src/components/counterparts/CounterpartDetails/useCounterpartDetails.tsx
caiomachado-monite: @costa-monite I like the idea, but how would you go about doing that? is that even possible to do? the main reference is in componentSettings

packages/sdk-react/src/components/counterparts/components/CreateCounterpartModal.tsx
costa-monite: I see that we rely on the componentSettings in that case, so everything should be fine

packages/sdk-react/src/components/counterparts/CounterpartDetails/CounterpartForm/CounterpartIndividualForm/CounterpartIndividualForm.tsx
```diff
@@ -77,12 +79,11 @@ export const CounterpartIndividualForm = ({
   /** Returns `true` if the form works for `update` but not `create` flow */
   const isUpdateMode = useMemo(() => Boolean(counterpart), [counterpart]);
```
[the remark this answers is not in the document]
caiomachado-monite: @costa-monite yeah, I personally don't like the current approach to define that, but I am open to suggestions, I didn't change that though because I didn't want to get too further into refactoring the whole thing, but let me see what I can think of and I will get back to you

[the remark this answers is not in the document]
caiomachado-monite: @costa-monite I removed the isUpdateMode completely because it seemed redundant to be honest, if the id exists, it means we're updating it because otherwise it wouldn't exist, great thinking though

packages/sdk-react/src/components/counterparts/CounterpartDetails/useCounterpartDetails.tsx
costa-monite: as I see CustomerType is defined in `counterparts/types.ts` so I think we can extend it like this:

```typescript
/**
 * Represents the type of a counterpart in the system.
 */
export type CustomerType = 'customer' | 'vendor';

/**
 * Array of available customer types, an array that should contain either customer, vendor, or both.
 * This array can't be empty and if only one option is passed, the customer type section will be hidden
 * and the default customer type will be the one passed.
 * It is set to undefined at component level but defaults to ['customer', 'vendor'] through componentSettings
 */
export type CustomerTypes = CustomerType[];
```

and for example in useCounterpartDetails.tsx we can use it like this:
```typescript
export type CounterpartsDetailsProps = {
  /** @see {@link CustomerTypes} */
  customerTypes?: CustomerTypes;
```

costa-monite (approved):
Thanks, huge work is done here 🚀

packages/sdk-react/src/components/counterparts/CounterpartDetails/useCounterpartDetails.tsx
caiomachado-monite: @costa-monite Nice! That's a great idea, thanks for the example, I went ahead and implemented it

PR #648, opened 2025-03-18
fix(proposal): SDK design refresh

marcperezmonite:


examples/with-nextjs-and-clerk-auth/src/app/(monite)/[[...path]]/page.tsx
```diff
@@ -33,10 +36,10 @@ export default function DefaultPage() {
           alignItems={'center'}
           sx={{ width: '100%', mb: '24px' }}
         >
-          <h1>Dashboard</h1>
+          <h1>Welcome, {user?.firstName}</h1>
```
Comment (costa-monite) [inline examples/with-nextjs-and-clerk-auth/src/app/(monite)/[[...path]]/page.tsx:39]: Might we here add the comma only if user.firstName is defined, to avoid the hanging comma state.
eg, until the user is defined we see just a "Welcome", after it will become "Welcome, {userName}"

packages/sdk-react/src/core/theme/mui-monite/index.ts
```diff
@@ -257,33 +257,33 @@ export const getTheme = (theme: ThemeConfig): ThemeOptions => {
...
+          // },
+          // '&.Monite-SummaryCard-StatusTypography-partially_paid': {
+          //   color: statusBackgroundColors.partially_paid,
+          // },
+          // '&.Monite-SummaryCard-StatusTypography-canceled': {
+          //   color: statusBackgroundColors.canceled,
+          // },
+          // '&.Monite-SummaryCard-StatusTypography-all': {
+          //   color: statusBackgroundColors.all,
+          // },
```
costa-monite, line 286: These and other commented-out blocks: do we need them or is it safe to remove them?

costa-monite (approved):
2 non crit comments and a TS error to fix and we're good to go. A huge work has been done here 🚀

examples/with-nextjs-and-clerk-auth/src/app/(monite)/[[...path]]/page.tsx
Comment (marcperezmonite) [reply examples/with-nextjs-and-clerk-auth/src/app/(monite)/[[...path]]/page.tsx:39]: Good catch!

PR #653, opened 2025-03-25
fix(proposal): Dropin unmounting

marcperezmonite:
Adds an `unmount()` method to Drop-in instances. This removes the DOM element where the monite app was mounter, as well as removes all event listeners automatically.


caiomachado-monite (approved):

packages/sdk-drop-in/monite-app-demo.html
```diff
@@ -16,44 +16,14 @@
     <title>Monite App Widget Label</title>
   </head>
   <body>
-    <monite-app
```
dinis-monite, line 19: We no longer have custom HTML element usage?

marcperezmonite: We do, but the preferred integration is to add it programmatically through: 

`dropin.create(component).mount(document.body);`

(that was changed in a different PR, but updating the Demo here to reflect it)

PR #660, opened 2025-03-31
feat(proposal): Add Document templates to Dropin

marcperezmonite:


.changeset/blue-horses-flash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@monite/sdk-drop-in': patch
+'@monite/sdk-react': patch
```
costa-monite, line 3: do we have to update the sdk-react version in this PR?

costa-monite (approved):
👍

.changeset/blue-horses-flash.md
marcperezmonite: I guess not, but ideally those two packages always have the same synced version. We could create a link between them so they always bump automatically, let's discuss separately.

PR #667, opened 2025-04-02
feat(DEV-14633): Implement PDF viewer

marcperezmonite:


packages/sdk-react/src/ui/FileViewer/FileViewer.tsx
```diff
@@ -37,38 +40,70 @@ export const FileViewer = ({
...
+      style={{
+        width: '100%',
+        height: '100%',
+        maxHeight: '100%',
+        border: 'none',
+      }}
+    >
+      <div
+        style={{
+          display: 'flex',
```
caiomachado-monite, line 62: Not blocking for me, but should we use these styles inline or use tailwind classes? I am ok with the inline approach as well for this specific component

costa-monite (approved):
🚀

dinis-monite (approved):

PR #680, opened 2025-04-10
feat(DEV-14646): Payables table prevent row click while processing OCR status

dinis-monite:
# Scope

Payables table: prevent row click (and disable hover effect) for rows of payables with OCR status in processing.

https://monite.atlassian.net/browse/DEV-14646

# Implementation

- changed the onRowClick prop of the DataGrid to trigger if status is not processing.
- added 'ocr-processing' className to rows that have OCR status in procesing and disable hover effects for that className.

# Steps to test

1. Open Payables table.
2. Click "add new bill" and submit a PDF file.
3. Check the first row, that shows "Processing file....", the row should not be clickable and have no hoverable effects.

### DIFF
--- a/.changeset/eighty-grapes-collect.md
+++ b/.changeset/eighty-grapes-collect.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+Payables table: prevent row click (and disable hover effect) for rows of payables with OCR status in processing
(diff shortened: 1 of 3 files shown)


caiomachado-monite (approved):
LGTM

costa-monite (approved):
Unrelated linter error that could be fixed here 👍

PR #685, opened 2025-04-14
fix(DEV-14870): repopulate the bank account values after submission

costa-monite:
**Changes**

*   **Fixed Field Reset on Submit:** Resolved an issue where Country and IBAN fields would reset after submitting the bank account form. This was achieved by:
    *   Ensuring the data transformation (`prepareBankAccountDataForPatch`) correctly uses the API mask.
    *   Explicitly resetting the React Hook Form state using `reset(formValues)` at the end of the `primaryAction` callback. This ensures the form's internal state reflects the submitted values even if subsequent effects modify the hook's local state.
*   **Improved Deletion Error Handling:** Added a `toast.error` notification in the `onError` handler for the `deleteBankAccountMutation` to provide user feedback if deleting a previous bank account fails (which can be expected during onboarding). Kept the `try...catch` block within `primaryAction` to allow the submission process to continue despite deletion errors

### DIFF
--- a/.changeset/modern-dots-sneeze.md
+++ b/.changeset/modern-dots-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+Fixed the issue with restoring values in the bank account onboarding after submission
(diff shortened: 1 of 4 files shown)


marcperezmonite:
Closing this in favor of #684. Let's keep in mind some of these features here thought, since we might need them in the future.

PR #697, opened 2025-04-22
feat(DEV-14965): Bank account creation hotfixes

caiomachado-monite:
- Added proper error handling on bank account creation when it fails displaying the actual error message instead of a generic one

- Updated invoice preview with bank account information (This is probably a temporary fix because we might replace the current preview with the new templates)

- Updated delete bank account modal visuals to be more like the figma version

- Fixed issue displaying undefined when the account was EUR, it was undefined because it was trying to read account number but EUR accounts only have IBAN

- Removed alert message and delete button from default bank account edit states

### DIFF
--- a/packages/sdk-react/src/components/receivables/components/BankAccountDeleteModal.tsx
+++ b/packages/sdk-react/src/components/receivables/components/BankAccountDeleteModal.tsx
@@ -1,14 +1,7 @@
 import { useRootElements } from '@/core/context/RootElementsProvider';
 import { t } from '@lingui/macro';
 import { useLingui } from '@lingui/react';
-import {
-  Button,
-  Dialog,
-  DialogActions,
-  DialogContent,
-  DialogTitle,
-  Divider,
-} from '@mui/material';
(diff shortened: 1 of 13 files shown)


dinis-monite (approved):

packages/sdk-react/src/components/receivables/components/BankAccountDeleteModal.tsx
```diff
@@ -35,13 +28,37 @@ export const BankAccountDeleteModal = ({
       maxWidth="sm"
       fullWidth
     >
-      <DialogTitle variant="h3">{t(i18n)`Delete bank account?`}</DialogTitle>
-      <DialogContent>{t(
-        i18n
-      )`You might need to update draft documents created with this bank account to ensure a correct payment.`}</DialogContent>
-      <Divider />
-      <DialogActions>
```
dinis-monite, line 43: Should we be changing the global styles of `DialogActions`?

costa-monite (approved):
🚀

PR #702, opened 2025-04-24
feat(DEV-14978): removed ssn_last_4 in favor of id_number

costa-monite:
## Problem
For US accounts (both Business and Individual), Stripe initially requests only the last 4 digits of the Social Security Number, but invariably asks for the full SSN later in the verification process. This creates a disjointed user experience where customers need to enter their SSN information twice.

## Changes
* Removed ssn_last_4 field from forms, validation schemas and mocks
* Created a `getIdentificationLabel` utility function in `helpers.ts` that returns the appropriate label based on country code
* Updated all relevant components to use dynamic labeling based on country


dinis-monite (approved):

caiomachado-monite (approved):
LGTM

PR #707, opened 2025-05-02
fix(DEV-15016): Edit Payable Counterpart without propagation

dinis-monite:
# Scope

Handle form submission without propagation in Counterpart forms. Add is_vendor filter to PayableDetailsForm and PayableDetailsInfo queries.

Jira tickets https://monite.atlassian.net/browse/DEV-15017 and https://monite.atlassian.net/browse/DEV-15016

# Implementation

- Added is_vendor filter to Counterpart queries in `PayableDetailsForm` and `PayableDetailsInfo`.
- Changed form submission in Counterpart forms to handle without propagation by correctly defining form submit button with form name.
- Removed not necessary onUpdate handler from `CounterpartDetails` component.

# Steps to test

For edit handler:

1. Open a Payable and edit;
2. On the Vendor/Counterpatt input, click the edit button;
3. Edit some data of the Counterpart (name, address, bank account, etc.), should not trigger the submit of the Payable save.

For is_vendor:

1. Have one Counterpart of type costumer called “Alice C” and one Counterpart of type vendor called “Bob V”.
4. upload a Payable with text that is OCRed to vendor “Alice C”.
5. open Payble, should not see indication of matching vendor to “Alice C”.
6. the input to select Counterpart should only show “Bob V”.

### DIFF
--- a/.changeset/green-worms-lay.md
+++ b/.changeset/green-worms-lay.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+Properly format Tag Categories labels
--- a/.changeset/nervous-cameras-burn.md
+++ b/.changeset/nervous-cameras-burn.md
@@ -1,0 +1,5 @@
+---
+'@monite/sdk-react': patch
+---
+
+Fix bug of Payable form submission on saving of the Payable Vendor form.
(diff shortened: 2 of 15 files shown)


caiomachado-monite (approved):
LGTM, I had the same issue in receivables and the fix was to actually move the form outside of the other form, maybe this is the same here, I haven't seen the full code but if you think you can move the counterpart form outside of the main form, that would be a good fix as well, despite that, the code looks good and I have no blockers on this

costa-monite (approved):
🚀

dinis-monite:
FYI, I simplified the code, with no need for the custom `handleSubmitWithoutPropagation` function, by correctly using the form submit button with the form name. Check commit b5fffa7

PR #710, opened 2025-05-05
feat(DEV-13640): Payables Table non-hideable columns

dinis-monite:
# Scope

Make some Payables Table columns non-hideable. Add component setting and default values for non-hideable columns.

Jira ticket https://monite.atlassian.net/browse/DEV-13640

# Implementation

- Added `requiredColumns` key to `payables` object in Component Settings.
- Added a `DEFAULT_REQUIRED_COLUMNS` as a fallback value.
- Changed PayablesTable to mark columns as hideable according to `requiredColumns` setting.

# Steps to test

1. Go to Payables Table.
2. Click on the 3-dot menu of a column.
3. Click manage columns.
4. On the menu, should be able to hide all columns except for the "number" column


packages/sdk-react/src/components/payables/PayablesTable/types.ts
```diff
@@ -43,6 +43,7 @@ export interface MonitePayableTableProps {
   isShowingSummaryCards?: boolean;
   fieldOrder?: FieldValueTypes[];
   summaryCardFilters?: Record<string, PayablesTabFilter | null>;
+  requiredColumns?: [FieldValueTypes, ...FieldValueTypes[]];
```
costa-monite, line 46: nice one
could we do the same for `fieldOrder`?

dinis-monite: Yes, we can use the same type for fieldOrder. Will update PR

PR #713, opened 2025-05-07
fix(DEV-14992): Remove alert for bank account selection when account is assigned

caiomachado-monite:
Changes:
1- Added logic to remove error message when you select bank account
2- Readded the alert for when user tries to delete a default bank account
3- Fixed bank account selection displaying null when there is no bank name


dinis-monite (approved):

costa-monite (approved):
🚀

PR #736, opened 2025-06-04
feat(DEV-15294): AI Assistant table sorting

Marvelanda:


marcperezmonite:
@Marvelanda - All looks good! Only I found there seems to be an issue when sorting by amounts, any ideas?

image

PR #767, opened 2025-07-15
feat(DEV-15555): Allow configuration of invoice template settings page

caiomachado-monite:
This PR is an upgrade on the templateSettings componentSettings options.
It adds a new property called `availableTemplateIds` that is essentially a list of available template ids.
If the list is empty or undefined, it defaults to all available templates. This is important because in the future when we start having partners adding their own templates, this will facilitate the ability to filter down the list returned templates to be only the ones they provide or choose.


packages/sdk-react/src/components/templateSettings/hooks/useDocumentTemplatesApi.ts
```diff
@@ -19,13 +19,29 @@ export const useDocumentTemplatesApi = () => {
...
         ['invoice', 'receivable'].includes(template.document_type)
       ) || [];
 
+    if (
+      componentSettings?.templateSettings?.availableTemplateIds &&
+      componentSettings?.templateSettings?.availableTemplateIds?.length > 0
+    ) {
+      filtered = filtered.filter(
+        (template) =>
+          componentSettings?.templateSettings?.availableTemplateIds?.includes(
```
costa-monite, line 34: Here and in the length comparison above, nullish coalescing is no longer needed

costa-monite (approved):
👍

PR #832, opened 2025-09-16
feat(DEV-16053): Add Receipts Inbox and ResponsiveGrid component

dinis-monite:
# Scope

Implemented Expense Management Receipts Inbox screen with a listing of receipts.
Added a reusable ResponsiveGrid component to show cards in a grid, with infinite loading.

Jira task: https://monite.atlassian.net/browse/DEV-16053

# Implementation

- Created ResponsiveGrid component that shows cards in a responsive grid, with infinite scroll and filtering capabilities.
- Implemented ReceiptsInbox component with filtering, infinite scrolling, and receipt display as cards.
- Enhanced ImageFileViewer component with customizable objectFit property.
- Implemented TabBar component according to new design.
- Added shadcn Separator component.
- Fixed "danger" color variables and updated its uses on other places.

# Remarks

Currently, the text search in Receipts inbox is searching by the `receipt.document_id`, it should be refactored to search by `receipt.merchant_name`, but API is not yet implemented.

Currently, the query for receipts is not filtering for receipts by the current user, but it should. Is pending API implementation of the filter.

# Steps to test

- On the "Expenses" page, click the "archive" icon button on the top right.
- A fullscreen dialog title "Receipt inbox" should open.
- It should load a listing of all receipts, with infinite scroll to load more.
- There should be a text input to search receipts by document_id.
- Clicking on one of the tabs, changes the listing to filter by the receipt state (matched/unmatched).

## Summary by CodeRabbit

- New Features
  - Added Receipts Inbox to Expenses with tabs (All/Matched/Unmatched), debounced search by document ID, infinite scrolling, receipt previews, amounts, dates, and loading skeletons. Includes inbox email address with copy action and a new header button to open the inbox.
  - Image viewer now supports configurable object-fit modes (contain, cover, fill).
- Style
  - Updated danger color palette tokens and refined error-state colors for consistency.
- Chores
  - Added a UI dependency to support new components.
  - Updated translations with new inbox and grid-related strings.


packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { i18n } = useLingui();
+  const { locale } = useMoniteContext();
+  const { formatCurrencyToDisplay } = useCurrencies();
+
+  const userDisplayName = useMemo(() => getUserDisplayName(user || {}), [user]);
+
+  const isMatched = !!receipt.transaction_id;
+  const isImageFile =
+    receipt.file_url &&
+    getMimetypeFromUrl(receipt.file_url) !== 'application/pdf';
```
costa-monite, line 28: We have some mime types defined and used scattered throughout the codebase eg:
```
export const OCR_SUPPORTED_FORMATS = [
  'application/pdf',
  'image/jpeg',
  'image/png',
];
```
might be a good opportunity to add enum/const/types to centralize these usages

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const formattedDate = i18n.date(receipt.created_at, locale.dateFormat);
+  const formattedAmount =
+    receipt.total_amount && receipt.currency
+      ? formatCurrencyToDisplay(receipt.total_amount, receipt.currency)
+      : '\u00A0';
```
costa-monite, line 34: `const NBSP = '\u00A0'` ?

packages/sdk-react/src/components/expenses/Expenses.tsx
```diff
@@ -1,22 +1,53 @@
...
+              </Button>
+            </div>
+          }
+        />
       </div>
       <div className="mtw:flex-1 mtw:min-h-0">
         <UserTransactionsTable />
       </div>
+      {receiptsInboxOpened && (
+        <Dialog
```
caiomachado-monite, line 44: I know this doesn't fall on you specifically, but I do think we should update this Dialog component to use the shadcn one, in the components folder we have also DialogHeader, DialogFooter and etc which is not necessary, maybe this can be a future enhancement or if it's not that big of a lift, we could update it here as well, what do you think?

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </span>
+            <span className="mtw:text-end">{formattedIssuedDate}</span>
+          </div>
+        </div>
+      </CardContent>
+    </Card>
+  );
+};
+
+export const ReceiptCard = memo(ReceiptCardComponent);
```
caiomachado-monite, line 97: I like the thought behind memoizing the component, however, I do think that it's overengeneering in this case because the component is just plain html pretty much, there's no expensive computation or anything that would justify avoid rerendering it, if we have a list of 1000 receipts, we would have 1000 memoized components, in this case we would be hurting performance more than helping it, because memoizing comes with a cost, what do you think? Again, I do like this approach and I think it's good, but maybe not for this specific case

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptsInbox.tsx
```diff
@@ -0,0 +1,280 @@
...
+  const receiptsEmailAddress = mailboxexData?.data?.find(
+    (mailbox) => mailbox.related_object_type === 'receipt'
+  )?.mailbox_full_address;
+
+  const handleReceiptClick = useCallback(
+    (
+      receipt: components['schemas']['ReceiptResponseSchema'],
+      index: number
+    ) => {
+      console.log('Clicked receipt:', receipt, 'at index:', index);
```
caiomachado-monite, line 145: just to keep track of this console.log, I know this is probably under construction still

packages/sdk-react/src/components/expenses/ReceiptsInbox/types.ts
caiomachado-monite, line 1: I would move this types file to the root of expenses folder, seems to me that this is more generic for expenses than focused on ReceiptsInbox, if this is focused on ReceiptsInbox, then we could move these types directly in the component where it's being used and avoid creating a separate file for a few types

packages/sdk-react/src/components/expenses/hooks/useReceipts.ts
caiomachado-monite, line 1: I would break both of these hooks into separate files for separation of concern

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
dinis-monite: I agree with centralizing the mime type strings into a enum/const/type. But I prefer to create a task for that, to avoid overloading the Expenses Management feature branch with non-feature changes, to easy the final merge.

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
dinis-monite: Yes. But can't use the HTML entity in JS, need to use the unicode code for that

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptCard.tsx
dinis-monite: Yeah, I also questioned some of the optimizations with memoizing.
That is a good point, this component here is mainly plain HTML, wouldn't benefit much with memoizing.
Will remove memoization here. Will update PR

packages/sdk-react/src/components/expenses/ReceiptsInbox/ReceiptsInbox.tsx
dinis-monite: Yes, this will be implemented in the next task. Hence the TODO code comment.

packages/sdk-react/src/components/expenses/ReceiptsInbox/types.ts
dinis-monite: You are right. Since these types (and consts) are only used in ReceiptsInbox, they can be kept there. Updated PR

costa-monite (approved):
👍

PR #833, opened 2025-09-16
feat(DEV-16270): Vat rate and recurrence fixes

caiomachado-monite:
Main changes:
1- Moved issue at field to before the start and end dates for recurrence
2- Reverted the change for the nonVatSupported fields, turns out the issue was actually on the entity we were using and that change was not needed
3- Fixed bug with recurrence not allowing the invoice to be created even when it was turned off
4- Changed the overview details for an invoice to only show accordion when there's something to show, otherwise hide it
5- Added a defaultMonth for the date picker fields, previously it was not showing the correct selected date when editing a recurrence

## Summary by CodeRabbit

- New Features
  - Added “Issue mode” selection (first day/last day) for recurring invoices; date pickers adapt to the selected mode, open to relevant months, and reset when mode changes.
  - Invoice items and preview respect non-VAT flows when applicable.
  - Overview UI now hides empty Linked Documents and Reminders sections and shows badges only when data exists.

- Bug Fixes
  - Fixed VAT rate field not displaying correctly when creating an invoice.
  - Recurrence toggle clears related errors and resets dates; validation runs only when recurrence is enabled.

- Chores
  - Added patch-level changeset for SDK packages.
  - Updated localization reference metadata without changing translations.

### DIFF
--- a/packages/sdk-react/src/components/receivables/components/RecurrenceFormContent.tsx
+++ b/packages/sdk-react/src/components/receivables/components/RecurrenceFormContent.tsx
@@ -43,7 +43,7 @@ export const RecurrenceFormContent = ({ isUpdate = false }) => {
                                 setValue('recurrence_start_date', undefined);
                                 setValue('recurrence_end_date', undefined);
                             }}
-                            defaultValue={field.value}
+                            value={field.value}
                         >
                             <FormControl>
                                 <SelectTrigger className="mtw:w-full">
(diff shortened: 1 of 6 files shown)


packages/sdk-react/src/components/receivables/components/RecurrenceFormContent.tsx
```diff
@@ -47,6 +77,7 @@ export const RecurrenceFormContent = ({ isUpdate = false }) => {
                                     shouldDisableButton={isUpdate}
                                     startMonth={new Date()}
                                     endMonth={new Date(2050, 11)}
+                                    {...(isUpdate && field.value ? { defaultMonth: new Date(field.value) } : {})}
```
costa-monite, line 80: Here and on line 106 we create new Date objects on every render when isUpdate and field.value are true, wont it be better to memoize the default date?

caiomachado-monite: @costa-monite we can't memoize it, the value depends on the render method of the calendar component, we need to know the field.value to get the correct date, if we memoize it to new Date it would be incorrect

